PR #2, opened 2017-07-11
change default connect timeout from 10ms to 10s

hjiebsco:
Noticed frequent Folio permission retrieval timeout in my AWS environment. It turns out the default 10ms is way below the original default value 60s. See http://vertx.io/docs/apidocs/io/vertx/core/net/ClientOptionsBase.html#DEFAULT_CONNECT_TIMEOUT. I modified the value from 10ms to 10s and do not see the timeout error again.

### DIFF
--- a/ModuleDescriptor.json
+++ b/ModuleDescriptor.json
@@ -1,10 +1,10 @@
 {
-  "id": "authtoken-module",
+  "id": "authtoken-module-0.5.0",
   "name": "authtoken",
   "provides": [
     {
       "id": "authtoken",
-      "version": "0.0.1",
+      "version": "1.0",
       "handlers" : [
         {
           "methods" : [ "POST" ],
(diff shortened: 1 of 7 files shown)


hjiebsco:
BTW, notice the different time unit between options.setConnectTimeout(**ms**) and options.setIdleTimeout(**s**)

julianladisch:
Looks good to me and may be merged. @kurtnordstrom ?

shale99:
do we need a 10s timeout for a connection - maybe 5s? this should be really quick as its just the connect part

julianladisch:
The 10s timeout is a default in case property `perm.lookup.timeout` has not been set:
https://github.com/folio-org/mod-authtoken/blob/0c73aad7e30e32c96bf4122452422164c8496009/src/main/java/org/folio/auth/authtoken_module/MainVerticle.java#L70

PR #13, opened 2017-11-09
FOLIO-787 Add nodeId to DeploymentDescriptor

dcrossleyau:


dcrossleyau:
Recent Okapi fails when it is missing in this deployment case.

wafschneider:
AFAIK, that is only true if you POST the deployment descriptor to the `/_/discovery/modules` endpoint -- if you post to `/_/deployment/modules` the nodeId is assumed to be the particular Okapi instance you are talking to.

A minor quibble, to be sure :-)

dcrossleyau:
Thanks for that clarification. It arose when i used mod-notes/run.sh to verify that these new DeploymentDescriptors were okay. That script does do POST to discovery.

I presume that it is okay for nodeId to be there, just is not used in that latter case.

PR #32, opened 2018-09-25
MODAT-35, MODAT-36: Fail startup if the required algorithm is not available. Fix key byte length.

julianladisch:


julianladisch:
WIP because:

DO NOT MERGE until https://issues.folio.org/browse/MODAT-36 has been fixed.

julianladisch:
MODAT-36 is also fixed and included in this pull request.

PR #47, opened 2019-03-15
MODAT-43 Bump up token perm cache from 10s to 60s

hjiebsco:


hjiebsco:
I am going to merge it. Hopefully nothing will break.

PR #50, opened 2019-06-05
Modat 46 duplicate x okapi permissions

adamdickmeiss:


adamdickmeiss:
Even the userPermissions have duplicates!

hjiebsco:
Are duplicates from database?

adamdickmeiss:
I don't know.. It might be that the permissions indeed return duplicate entries. It needs further investigation. Since those are part of a HTTP response body they do not impose an immediate thread :-)

PR #53, opened 2019-07-12
Make extra permissions work, fix cache key calculation MODAT-47

adamdickmeiss:
The URLEncoder.encode usage is OK and results in what appears
to be correct expanded permissions for extra-permission. The
extra permissions caching, however, is disabled by this and should
be fixed.

### DIFF
--- a/src/main/java/org/folio/auth/authtokenmodule/impl/ModulePermissionsSource.java
+++ b/src/main/java/org/folio/auth/authtokenmodule/impl/ModulePermissionsSource.java
@@ -32,14 +32,14 @@ public class ModulePermissionsSource implements PermissionsSource, Cache {
   private final Logger logger = LoggerFactory.getLogger("mod-auth-authtoken-module");
   private final HttpClient client;
   private LimitedSizeMap<String, CacheEntry> cacheMap;
-  private boolean cacheEntries;
+  private boolean cacheEnabled;
   private final String keyPrefix;
   private final int MAX_CACHE_SIZE = 250;
 
 
   public ModulePermissionsSource(Vertx vertx, int timeout, boolean cache) {
     //permissionsModuleUrl = url;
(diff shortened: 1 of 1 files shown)


marcjohnson-kint:
Does the disabling of the cache mean the long delay that is usually experienced only some of the time, now happens on each request?

hjiebsco:
Fixed the module permission expanding issue and restored the use of cache.

julianladisch:
Can you rename cacheEntries to cacheEnabled (or some other more self-explanatory name)?

adamdickmeiss:
Note that the caching is still sub optimal.. There should be two cache entries.. Not one.

The first (and existing one) should cache the result of getPermissionsForUser. So the key obviously depends on user and tenant.

The second is the call toexpandPermissions. It does NOT depend on the user. This is important. The cache entry for that one can have a long TTL, because it only depends on ModuleDescriptors enabled for the tenant. These rarely changes!!!

I think the code would also be clearer if it was a simple "intercept" of those calls. Since the former have a short TTL it makes the 2nd call being called at times where that is not necessary.

julianladisch:
@Adam Should we merge and use the solution of this pull request, and work on the 2 cache solution afterwards, or should we wait until the 2 cache solution is implemented?

julianladisch:
Adam answered on Slack that we can merge this fix and should work on a two cache solution afterwards.

julianladisch:
Created https://issues.folio.org/browse/MODAT-49 "Two caches for permissionsForUser and expandPermissions".

hjiebsco:
Got the nod from @skoczko, I will merge minimal code changes needed to fix this issue (https://github.com/folio-org/mod-authtoken/pull/56), so circ and patron tickets can be unblocked. Since we are going to refactor the cache in MODAT-49, this PR won't be needed then.

PR #61, opened 2019-09-13
Pass X-Okapi-Request-Id MODAT-50

adamdickmeiss:


julianladisch:
SonarClound's Quality Gate fails because the complexity of expandPermissions method is 91 which is above our limit of 30.
This can be ignored for this pull request because this pull request doesn't increase the complexity.
Working on expandPermissions method is a task for a separate issue.

adamdickmeiss:
Thanks for your approval.. I FWIW, went further and that's a lot of boring work.. Since the code overall has low coverage, any changes require new tests.

PR #67, opened 2019-11-24
MODAT-56 validate user deactivation when checking access token

hjiebsco:


src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -629,16 +645,36 @@ we generate a custom token (since we have that power) that
...
+    Future<PermissionData> retrievedPermissionsFuture;
+    if (finalUserId != null && !finalUserId.isEmpty()) {
+      Future<Boolean> activeUser = userService.isActiveUser(finalUserId, tenant, okapiUrl, userRequestToken, requestId);
+      retrievedPermissionsFuture = activeUser.compose(b -> {
+        if (b != null && b.booleanValue()) {
+          return usePermissionsSource.getUserAndExpandedPermissions(finalUserId, tenant, okapiUrl,
+              permissionsRequestToken, requestId, extraPermissions);
+        } else {
+          endText(ctx, 401, "Invalid token: user with id " + finalUserId + " is not active");
+          return Future.succeededFuture();
```
adamdickmeiss: Shoudln't this be Future.failedFuture() ?

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -629,16 +645,36 @@ we generate a custom token (since we have that power) that
     logger.debug("Getting user permissions for " + username + " (userId " +
             userId + ")");
     long startTime = System.currentTimeMillis();
-    Future<PermissionData> retrievedPermissionsFuture = usePermissionsSource
-      .getUserAndExpandedPermissions(userId, tenant, okapiUrl, permissionsRequestToken,
-        requestId, extraPermissions);
+
+    // Need to check if the user is still active
+    Future<PermissionData> retrievedPermissionsFuture;
+    if (finalUserId != null && !finalUserId.isEmpty()) {
```
adamdickmeiss: What is happening here? Empty or null? SQ says one.. IMHO, get rid of the isEmpty.

hjiebsco: Check to make sure user id is not null and is not empty (length = 0). Does that make sense?

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
hjiebsco: That makes sense. Fixed.

hjiebsco:
> The ModuleDescriptor should say that mod-authtoken now depend on users interface

Added users ad dependency.

PR #72, opened 2020-04-06
MODAT-62 - Issue with log4j configuration

craigmcnally:
## Purpose
Since moving to log4j2 (https://github.com/folio-org/mod-authtoken/pull/70) we've run into an issue with `log4j2Plugins.dat` colliding with those from 3rd party dependencies while shading.  

See: https://issues.apache.org/jira/browse/LOG4J2-673.
See: https://issues.folio.org/browse/MODAT-62

## Approach
Use a shade plugin transformer which merges the various `log4jPlugins.dat` files.

See https://github.com/edwgiz/maven-shaded-log4j-transformer

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -179,12 +179,21 @@
                     <Main-Verticle>org.folio.auth.authtokenmodule.MainVerticle</Main-Verticle>
                   </manifestEntries>
                 </transformer>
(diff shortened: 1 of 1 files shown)


julianladisch:
@hjiebsco I don't see any other solution for now. I've created https://issues.folio.org/browse/MODAT-63 as a reminder.

PR #73, opened 2020-05-08
MODAT-72 Expand module permission set

hjiebsco:
In OKAPI-837, we automatically convert module permissions to a permission set. Unfortunately mod-authtoken does not expand permission set to permissions for certain scenarios where no real user id is involved. For example during initial login or Okapi timer based system calls. This PR is to address this issue.


hjiebsco:
> Looks good. Just think you should bump the interface for mod-permissions with the expanded parameter

Updated mod-authtoken to depend on the new **permissions** interface

PR #75, opened 2020-06-02
Update to Vert.x 3.9.1 MODAT-76

adamdickmeiss:


adamdickmeiss:
getAbs can not be fixed unless we switch to Vert.x Web client. Should I do that?

src/main/java/org/folio/auth/authtokenmodule/impl/ModulePermissionsSource.java
```diff
@@ -155,27 +151,23 @@ private void endRequest(HttpClientRequest req, String requestToken,
...
   private Future<JsonArray> expandPermissions(JsonArray permissions, String tenant,
     String okapiUrl, String requestToken, String requestId) {
-    Future<JsonArray> future = Future.future();
+    Promise<JsonArray> promise = Promise.promise();
     if (permissions.isEmpty()) {
-      future.complete(new JsonArray());
-      return future;
+      promise.complete(new JsonArray());
+      return promise.future();
     }
```
hjiebsco, line 171: Minor suggestion: the **promise** declaration can be moved down to save an extra line.
```
    if (permissions.isEmpty()) {
      return Future.succeededFuture(new JsonArray());
    }
    Promise<JsonArray> promise = Promise.promise();
```

pom.xml
```diff
@@ -38,12 +38,12 @@
     <dependency>
       <groupId>io.vertx</groupId>
       <artifactId>vertx-core</artifactId>
-      <version>3.5.4</version>
+      <version>${vertx.version}</version>
```
julianladisch, line 41: Instead of introducing ${vertx.version} consider using Bill of Materials (Maven BOM): https://github.com/vert-x3/vertx-stack#bills-of-materials-maven-bom
This also sets the netty and jackson version: https://github.com/vert-x3/vertx-dependencies/blob/master/pom.xml#L50-L53

src/main/java/org/folio/auth/authtokenmodule/impl/ModulePermissionsSource.java
adamdickmeiss: Yep

PR #78, opened 2020-06-24
Skip expanded system permissions

hjiebsco:
Please see https://issues.folio.org/browse/MODAT-79


hjiebsco:
> Reluctantly approve. IMHO this function and the work with those permissions seems now extremely complicated.

Hopefully the redesign of Auth in Q3 will improve this.

PR #88, opened 2020-12-17
Replace linear search optimized queue MODAT-82

adamdickmeiss:


src/main/java/org/folio/auth/authtokenmodule/LimitedSizeQueue.java
```diff
@@ -0,0 +1,41 @@
...
+    if (!super.equals(obj)) {
+      return false;
+    }
+    LimitedSizeQueue<K> fobj = (LimitedSizeQueue<K>) obj;
+    return fobj.maxSize == maxSize;
+  }
+
+  @Override
+  public int hashCode() {
+    return super.hashCode();
```
julianladisch, line 25:
```suggestion
    return super.hashCode() + 31 * maxSize;
```
Because maxSize is relevant for equals.

src/test/java/org/folio/auth/authtokenmodule/LimitedSizeQueueTest.java
```diff
@@ -0,0 +1,86 @@
...
+    System.out.printf("time = " + (end - start) / 1000000 + " ms");
+  }
+
+  @Test
+  public void testQueueCompare() {
+    LimitedSizeQueue<String> a1 = new LimitedSizeQueue<>(1);
+    LimitedSizeQueue<String> a2 = new LimitedSizeQueue<>(2);
+
+    Assert.assertNotEquals(a1, a2);
+    Assert.assertEquals(a1.hashCode(), a2.hashCode());
```
julianladisch, line 74:
```suggestion
    Assert.assertNotEquals(a1.hashCode(), a2.hashCode());
```
The hashCode implementation is bad if a1.equals(a2) but hashCode is the same.

src/test/java/org/folio/auth/authtokenmodule/LimitedSizeQueueTest.java
```diff
@@ -0,0 +1,86 @@
...
+    LimitedSizeQueue<Integer> q = new LimitedSizeQueue<>(2);
+    Assert.assertTrue(q.isEmpty());
+    q.add(1);
+    Assert.assertEquals(1, q.size());
+    Assert.assertTrue(q.contains(1));
+    q.add(2);
+    Assert.assertEquals(2, q.size());
+    Assert.assertTrue(q.contains(1));
+    Assert.assertTrue(q.contains(2));
+    // observe that putting 1 here does NOT put it in front of queuee
```
julianladisch, line 38: This should be mentioned in the Javadoc of LimitedSizeQueue.add

src/test/java/org/folio/auth/authtokenmodule/LimitedSizeQueueTest.java
```diff
@@ -0,0 +1,86 @@
...
+    for (int i = 0; i < count; i++) {
+      if (i < count - sz) {
+        Assert.assertFalse(q.contains(prefix + i));
+      } else {
+        Assert.assertTrue(q.contains(prefix + i));
+      }
+    }
+    long end = System.nanoTime();
+    System.out.printf("time = " + (end - start) / 1000000 + " ms");
+  }
```
julianladisch, line 66: This test should have `@Ignore("Manual performance test")` because there is no execution time assert. (The `contains` tests are not really needed because this is checked in the previous tests.)
MainVerticle.MAX_CACHED_TOKENS should be increased to make better use of this performance improvement.

src/test/java/org/folio/auth/authtokenmodule/LimitedSizeQueueTest.java
adamdickmeiss: There's nothing wrong with this. But can change hash implementation.

src/main/java/org/folio/auth/authtokenmodule/LimitedSizeQueue.java
adamdickmeiss: ok

src/test/java/org/folio/auth/authtokenmodule/LimitedSizeQueueTest.java
adamdickmeiss: Added it. Since the usage is that it checks with contains before adding, there's nothing assumed here.

src/test/java/org/folio/auth/authtokenmodule/LimitedSizeQueueTest.java
adamdickmeiss: How about I just remove the code entirely?

julianladisch: Removal is fine with me.

src/main/java/org/folio/auth/authtokenmodule/LimitedSizeQueue.java
```diff
@@ -0,0 +1,45 @@
...
+    return super.hashCode() + 31 * maxSize;
+  }
+
+  @Override
+  protected boolean removeEldestEntry(Map.Entry<K, Boolean> eldest) {
+    return size() > maxSize;
+  }
+
+  /**
+   * Adds key if not already in queue. If key exists, nothing is changed.
```
julianladisch, line 34:
```suggestion
   * Adds key if not already in queue. If key exists, its queue position doesn't change.
```

julianladisch (approved):
Looks good, thanks.

PR #89, opened 2021-02-27
log4j2 props, Vert.x 4.0.2 MODAT-98

adamdickmeiss:
Some updates to log4j2 calls as well with {}-formatting.

### DIFF
--- a/src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
+++ b/src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
@@ -33,6 +33,7 @@ import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.config.Configurator;
 import org.folio.auth.authtokenmodule.impl.DummyPermissionsSource;
 import org.folio.auth.authtokenmodule.impl.ModulePermissionsSource;
+import org.folio.okapi.common.logging.FolioLoggingContext;
 
 /**
  *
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -185,7 +186,7 @@ private TokenCreator lookupTokenCreator(String passPhrase) throws JOSEException
 
   private void handleSignEncryptedToken(RoutingContext ctx) {
     try {
-      logger.debug("Encrypted token signing request from " + ctx.request().absoluteURI());
+      logger.debug("Encrypted token signing request from {}", () -> ctx.request().absoluteURI());
```
hjiebsco, line 189: Not consistent. Other logging statements do not use `() ->`

hjiebsco (approved):

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
adamdickmeiss: Right. Not elegant. Go through the rest or revert this one?

hjiebsco: I vote changing this one to be consistent with others. The use of `{}` already makes the string construction delayed in log4j2. See https://logging.apache.org/log4j/2.x/manual/api.html

adamdickmeiss: ok.

hjiebsco (approved):

PR #99, opened 2021-12-02
MODAT-109: implement new token types

steveellis:
The overall goal here is to prepare mod-authtoken for changes based on the [new token architecture](https://docs.google.com/document/d/1K_QdgnOo2wOSfY-rQ8phOD6nCO_3jvdAnEG0BEqtnjU/edit?usp=sharing). This doesn't complete the work, but considering the extent of the changes, we're breaking things up into stages.

The goals for this stage were as follows:
* Create separate classes for each new token type and existing token types.
* Make each token type be responsible for its own validation.
* Provide simple API for validating tokens, whatever the token type and context.
* Make the token lifecycle more clear, from creation to validation, to request completion.
* Ensure that the code can be merged to master without any breaking changes (all existing unit tests pass).
* Clean up `MainVerticle.handleAuthorize` wherever possible to make the code more readable.

Much of this is handled through the new `Token` class, which is an abstract class with one method that inheritors need to implement for validation. Token lifecycle is now easier to follow since all tokens are created their class constructors. Validation itself is handled through a static `Token.validate` method that makes use of `Future`'s `onFailure` and `onSuccess` to take care of the new async nature of validation. `Token.validate` is a convenience method that parses the token and calls each type's `validate` method.

Validation now needs to be async since some tokens will require IO for validation.


src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,262 @@
...
+import java.text.ParseException;
+
+/**
+ * An abstract class that all token types should extend, with one required method
+ * that all must implement to handle their validation logic. This class also exposes
+ * a handful of static methods which provide convenient ways of working with tokens,
+ * including a method to parse and validate a token in one step.
+ */
+public abstract class Token {
+  private static final String CALLING_MODULE_HEADER = "X-Okapi-Calling-Module";
```
adamdickmeiss, line 22: This header has been there for a long time. Okapi doesn't set it. Never has. I think we can remove it (and handling for it).

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,262 @@
...
+  private boolean usesDummyPermissionsSource;
+  protected static final String UNDEFINED_USER_NAME = "UNDEFINED_USER__";
```
adamdickmeiss, line 24: This also bugs me as a hack. Not your fault though. I wished it could be removed.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * Some tokens require that the dummy permissions source be used. This will return
+   * true when that is the case.
+   * @return True if the dummy permissions source should be used.
+   */
+  public boolean shouldUseDummyPermissionsSource() {
+    // Dummy tokens are not the only type of tokens that require the use of this
+    // so checking the token type alone isn't enough. We also have to check the sub claim.
+    if ((claims.getBoolean("dummy") != null &&
+         claims.getBoolean("dummy"))
```
adamdickmeiss, line 85:
```suggestion
    if (Boolean.TRUE.equals(claims.getBoolean("dummy"))
```

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,262 @@
...
+      || claims.getString("sub").startsWith(Token.UNDEFINED_USER_NAME)) {
```
adamdickmeiss, line 86: The UNDEFINED_USER_NAME looks redundant..set in DummyToken constructor, and dummy=true anyway.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+   * Will add the calling module if the request headers require it.
+   * @param requestHeaders The headers in the http context where the token is being provided.
+   */
+  public void tryAddCallingModule(MultiMap requestHeaders) {
+    if (requestHeaders.contains(CALLING_MODULE_HEADER)) {
+      claims.put("calling_module", requestHeaders.get(CALLING_MODULE_HEADER));
+    }
+  }
```
adamdickmeiss, line 112: As mentioned, can be removed as nothing sets this header.

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenType.java
```diff
@@ -0,0 +1,9 @@
+package org.folio.auth.authtokenmodule.tokens;
+
+public final class TokenType {
+  public static final String REFRESH = "refresh";
+  public static final String ACCESS = "access";
+  public static final String API = "api";
+  public static final String MODULE = "module";
+  public static final String DUMMY = "dummy";
+}
```
adamdickmeiss, line 9: Could also have been const public static member of each Token class implementation to avoid keeping things "in-sync"

adamdickmeiss (commented):
I'd like to take the branch for a spin the vagrant testing box.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
julianladisch: Yes, not used anywhere: https://github.com/search?q=org%3Afolio-org+X-Okapi-Calling-Module&type=code

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
```diff
@@ -0,0 +1,29 @@
+package org.folio.auth.authtokenmodule.tokens;
+
+/**
+ * When a token is provided and it hasn't passed validation, for any reason,
+ * a TokenValidationException will be present. Token validation exceptions have
+ * an opinion about what http response code should be returned to clients.
+ */
+public class TokenValidationException extends Exception {
+  public int httpResponseCode;
```
julianladisch, line 9: Can you make it private and add a getter?
Can you make it final?

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public TokenValidationException(String message, int responseCode) {
+    super(message);
+    httpResponseCode = responseCode;
```
julianladisch, line 13: Can you rename `responseCode` to `httpResponseCode` and use `this.httpResponseCode = httpResponseCode` in this and next method?

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -265,50 +256,50 @@ private void handleDecodeEncryptedToken(RoutingContext ctx) {
...
-          return;
-        }
-        if (!res.result()) {
-          endText(ctx, 401, "Invalid refresh token");
+  
+      String encryptedJWE = requestJson.getString("refreshToken");
+      Future<Token> tokenValidationResult = Token.validate(encryptedJWE, tokenCreator, ctx.request());
+
+      tokenValidationResult.onFailure(h -> {
+        var e = (TokenValidationException)h;
```
julianladisch, line 279: There is no guarantee that all Exceptions returned by a Future are a TokenValidationException.
```
if (h instanceof TokenValidationException) {
   ...
} else {
   ...
}
```

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -265,50 +256,50 @@ private void handleDecodeEncryptedToken(RoutingContext ctx) {
...
+        logger.error("Invalid refresh token: {}", e.toString());
+        endText(ctx, e.httpResponseCode, "Invalid token");
+      });
+
+      tokenValidationResult.onSuccess(h -> {
+        Token token = (Token)h;
```
julianladisch, line 285: No cast needed because of `Future<Token>`.

julianladisch: `tokenValidationResult.onSuccess(token -> {`

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,262 @@
...
+   * @param token A string version of the token to test.
+   * @return True if the token is encrypted, otherwise false.
+   * @throws TokenValidationException
+   */
+  public static boolean isEncrypted(String token) throws TokenValidationException {
+    // This is based on how JOSEObject parses encrypted tokens. It checks the number of "parts"
+    // which are equivalent to the number of "." separators in the token. If there are 5 parts
+    // it is considered to be encrypted. If there are not 5 it is rejected from JWE parsing.
+    // Our unencrypted tokens have 3 parts (and 2 separators).
+    int parts = StringUtils.countMatches(token, ".");
```
julianladisch, line 160: No need to add the complete commons-lang3 library - you can easily make a for loop to count the dots (optionally: put the loop into a `countDots(String) method).
Since version 3.4 `StringUtils.countMatches(token, ".")` can be replaced by `StringUtils.countMatches(token, '.')`.

adamdickmeiss (changes requested):
I've tried mod-authtoken in folio testing Vagrant box. Something goes wrong WRT dummy permissions and the dreaded `UNDEFINED_USER`. See below.. My theory is that in some place the dummy user use used in a wrong place. See if you can get rid of it entirely ..
```
14:20:19 [882563/bl-users] [diku] [] [mod-authtoken] ERROR MainVerticle         Invalid token: TokenValidationException: Invalid token class org.folio.auth.authtokenmodule.BadSignatureException
14:20:19 [882563/bl-users] [diku] [] [mod-authtoken] ERROR MainVerticle         Invalid token
14:20:23 [436211/bl-users;440152/authn;436676/perms] [diku] [] [mod-authtoken] ERROR MainVerticle         [][] (user permissions) nor [] (module permissions) do not contain perms.permissions.get
14:20:23 [436211/bl-users;440152/authn;436676/perms] [diku] [] [mod-authtoken] ERROR MainVerticle         Access requires permission: perms.permissions.get
14:20:23 [436211/bl-users;440152/authn] [diku] [] [mod-authtoken] ERROR ulePermissionsSource Error expanding ["SYS#mod-users-bl-7.2.0-SNAPSHOT.107#/bl-users/login#[POST]"]: Expected 200, got result 403 : Access requires permission: perms.permissions.get
14:20:23 [436211/bl-users;440152/authn] [diku] [] [mod-authtoken] ERROR MainVerticle         Unable to retrieve permissions for UNDEFINED_USER__10.0.2.15:59882__2021-12-03T14:20:23.533+0000: Expected 200, got result 403 : Access requires permission: perms.permissions.get request took 92 ms
14:20:23 [436211/bl-users;440152/authn] [diku] [] [mod-authtoken] ERROR MainVerticle         Unable to retrieve permissions for user with id 'null': Expected 200, got result 403 : Access requires permission: perms.permissions.get
14:20:31 [387542/bl-users;570262/authn;783313/perms] [diku] [] [mod-authtoken] ERROR MainVerticle         [][] (user permissions) nor [] (module permissions) do not contain perms.permissions.get
14:20:31 [387542/bl-users;570262/authn;783313/perms] [diku] [] [mod-authtoken] ERROR MainVerticle         Access requires permission: perms.permissions.get
```

steveellis:
@julianladisch Thanks for your comments. I'll make those changes soon. And @adamdickmeiss thanks for alerting me to the problems after deploying it to the testing env. I was able to get that going, was able to repo, and then trace it back to my refactor of this [bit of code](https://github.com/folio-org/mod-authtoken/blob/6d0438ab6de9638322f88bb8c62174ae1edca183/src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java#L536). I didn't like that it was doing two different things (checking for the header/token match) and then doing the assignment, but I missed that the assignment needed to happen with the two conditions being true. The header/token match check now happens in Token.validate.

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -456,316 +432,272 @@ private void handleAuthorize(RoutingContext ctx) {
...
+      // give what is here and in the existing code called the finalUserId a value
+      // in this specific case of the tokenUserId having a value _and_ the X-Okapi-User-Id
+      // being null. I'd likek to better understand the circumstances under which these
+      // two variables can have this state and what it means.
+      String temporaryUserId = userId;
+      String tokenUserId = token.getClaims().getString("user_id");
+      if (tokenUserId != null && userId == null) {
+          temporaryUserId = tokenUserId;
+      }
+      final String finalUserId = temporaryUserId;
```
julianladisch, line 493:
```
String headerUserId = ctx.request().headers().get(XOkapiHeaders.USER_ID);
```

```
String tokenUserId = token.getClaims().getString("user_id");
final String userId = headerUserId != null ? headerUserId : tokenUserId;
```

For me this is more readable.

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -456,316 +432,272 @@ private void handleAuthorize(RoutingContext ctx) {
...
+      logger.debug("Validated token of type: {}", token.getClaims().getString("type"));
+
+      String username = token.getClaims().getString("sub");
+
+      // TODO I believe this block solves the issue that Adam found with functional testing.
+      // My approach of just making userId be equal to the X-Okapi-User-Id didn't
+      // give what is here and in the existing code called the finalUserId a value
+      // in this specific case of the tokenUserId having a value _and_ the X-Okapi-User-Id
+      // being null. I'd likek to better understand the circumstances under which these
+      // two variables can have this state and what it means.
```
julianladisch, line 487: Okapi uses that same logic: https://github.com/folio-org/okapi/blob/v4.9.0/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java#L280

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Will remove.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: @adamdickmeiss you mean the whole `UNDEFINED _USER__` thing? My sense is that this is needed to handle the pre-auth state (before login). But I haven't dug into it much. Just tried to keep my code from breaking the existing functionality. But I agree it doesn't smell right.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Agree. Why not `token.split("\\.").length`? For some reason I was thinking about counting a character...

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
steveellis: Agree the one liner is more readable. Still find it somewhat troubling that we don't have a single source of truth for the userId. Seems fine that that source of truth be in the header _or_ the token, but what bothers me (and led to a difficult bug because I assumed that it would be in the header) was that it wasn't always defined in the header if it was known.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
julianladisch:
```suggestion
    if (Boolean.TRUE.equals(claims.getBoolean("dummy"))
```

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
julianladisch: `token.split("\\.").length` is a nice hack - it creates strings and then throws them away. The final result is correct, the code is somewhat misleading.

```
    long dots = token.chars().filter(c -> c == '.').count();
```

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -456,316 +437,265 @@ private void handleAuthorize(RoutingContext ctx) {
...
       }
-    }).compose(expandedPermissions -> {
-      expandedSystemPermissions.addAll(expandedPermissions);
-      // skip expanded system permissions
-      JsonArray extraPermsMinusSystemOnes = new JsonArray();
-      extraPermissions.forEach(it -> {
-        if (!((String)it).startsWith(PermService.SYS_PERM_PREFIX)) {
-          extraPermsMinusSystemOnes.add(it);
+      Future<PermissionData> retrievedPermissionsFuture = activeUser.compose(b -> {
+        if (b != null && b.booleanValue()) {
```
julianladisch, line 594:
```suggestion
        if (Boolean.TRUE.equals(b)) {
```
And even
```
        if (TRUE.equals(b)) {
```
is possible after `import static java.lang.Boolean.TRUE;`

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class TokenValidationException extends Exception {
+  /**
+   * Returns the HTTP response code to return when this exception is encountered.
+   * @return The HTTP response code.
+   */
+  public int getHttpResponseCode() {
+    return httpResponseCode;
+  }
+  private int httpResponseCode;
```
julianladisch, line 16: I can find variables more easily if they are all before the first method. https://dev.folio.org/guidelines/contributing/#coding-style , https://www.oracle.com/java/technologies/javase/codeconventions-fileorganization.html#1852

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public TokenValidationException(String message, int httpResponseCode) {
+    super(message);
+    this.httpResponseCode = httpResponseCode;
+  }
+
+  public TokenValidationException(String message, Exception cause, int responseCode) {
```
julianladisch, line 23: `responseCode` -> `httpResponseCode`

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
```diff
@@ -0,0 +1,36 @@
...
+    super(message, cause);
+    httpResponseCode = responseCode;
+  }
+
+  @Override
+  public String toString() {
+    if (getCause() != null)
+      return String.format("TokenValidationException: %s %s", getMessage(), getCause().getClass());
```
julianladisch, line 31: For readability curly braces around a single Java statement are never omitted. https://dev.folio.org/guidelines/contributing/#coding-style , https://www.oracle.com/java/technologies/javase/codeconventions-statements.html#15395

pom.xml
```diff
@@ -236,4 +241,4 @@
       </plugin>
     </plugins>
   </build>
-</project>
+</project>
```
julianladisch, line 239: Can you configure your IDE to not remove the new line at the end of the file?

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,248 @@
...
+        throw new TokenValidationException("User id in header does not equal user id in token", 403);
+
+    } catch (TokenValidationException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new TokenValidationException("Unexpected token validation exception", e, 500);
+    }
+  }
+
+  private static Token parse(String sourceToken, TokenCreator tokenCreator) throws TokenValidationException {
```
julianladisch, line 199: Can you move this method into a `TokenFactory` class and rename it to `create`?

src/main/java/org/folio/auth/authtokenmodule/tokens/ApiToken.java
```diff
@@ -0,0 +1,57 @@
...
+
+  protected Future<Token> validate(HttpServerRequest request) {
+    try { 
+      validateCommon(request);
+    } catch (TokenValidationException e) {
+      return Future.failedFuture(e);
+    }
+
+    // TODO Validate the API token by checking that it exists in storage,
+    // hasn't been revoked, etc.
```
julianladisch, line 53: How can the ApiToken access storage if it only gets HttpServerRequest as argument?

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
steveellis: Think this is done, no? What am I missing?

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
steveellis: Done

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
steveellis: I'm fine with this if it is what the standard is on the project. Personally I like getters and privates to be grouped.

pom.xml
steveellis: Done

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,248 @@
...
+      throw new TokenValidationException(invalidTokenMsg, b, 401);
+    }
+
+    String tokenType = claims.getString("type");
+    if (tokenType == null)
+      throw new TokenValidationException("Token has no type", 400);
+
+    switch (tokenType) {
+      case AccessToken.type:
+        token = new AccessToken(sourceToken, claims);
```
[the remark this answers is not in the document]
steveellis: Agree it would be less code but then callers have to deal with the potential of a null when a token has no type. Felt it was better to have this method handle that. I was imagining a situation when we would have some tokens "out there" that didn't have types and have a clear message when that occurs.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: I had that originally since it is the more canonical abstract factory pattern but then I thought it might be better to make this method private since validation is the only case when tokens will need to be created with these constructors. In other words there's no case when the factory would need to be used and we don't want to encourage it. Let me know what you think.

src/main/java/org/folio/auth/authtokenmodule/tokens/ApiToken.java
steveellis: One way to handle this would be to create a class called `TokenValidationContext` or similar and make the `HttpServerRequest` and any other properties like a db connection (and anything else we might need) be members. Then make this class the sole argument to Token.validate.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Think we need `UNDEFINED_USER_NAME` here. I had the same thought, but I believe it doesn't work.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: This should be done. Think Adam beat you to it.

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
steveellis: Cool! Added it.

src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java
steveellis: Oh I see it now. Done.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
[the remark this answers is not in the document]
julianladisch: A constructor never returns null. "The value of a class instance creation expression is a reference to the newly created object of the specified class." https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.9.4

    switch (tokenType) {
      case AccessToken.type:
        return new AccessToken(sourceToken, claims);
      case RefreshToken.type:
        return new RefreshToken(sourceToken, claims);
      case ApiToken.type:
        return new ApiToken(sourceToken, claims);
      case DummyToken.type:
        return new DummyToken(sourceToken, claims);
      case ModuleToken.type:
        return new ModuleToken(sourceToken, claims);
      default:
        throw new TokenValidationException("Unable to parse token", 400);
    }

src/main/java/org/folio/auth/authtokenmodule/tokens/ApiToken.java
julianladisch: Yes.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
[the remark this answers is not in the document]
steveellis: You are right. For some reason I was focused on the type being null, but I check for that above.

src/test/java/org/folio/auth/authtokenmodule/TokenTest.java
```diff
@@ -0,0 +1,73 @@
...
+    String tokenMissingTenantClaim =
+      "{\"iat\":1637696002,\"sub\":\"test-username\",\"user_id\":\"007d31d2-1441-4291-9bb8-d6e2c20e399a\",\"type\":\"access\"}";
+    String key = System.getProperty("jwt.signing.key");
+    var tokenCreator = new TokenCreator(key);
+    String source = tokenCreator.createJWTToken(tokenMissingTenantClaim);
+
+    Future<Token> result = Token.validate(source, tokenCreator, null);
+
+    assert(result.failed());
+    result.onFailure(e -> {
```
julianladisch, line 54: result.onFailure is not called until the future is completed. Therefore the unit test might end before it is called and the containing tests are not executed.
In this case the future is complete, but the construction should be avoided because code changes might change that in future.

    Future<Token> future = Token.validate(source, tokenCreator, null);

    assertThat(future.failed(), is(true));
    assertThat(future.cause(), is(instanceOf(TokenValidationException.class)));
    assertThat(((TokenValidationException) future.cause()).getHttpResponseCode(), is(500));

Or

    Token.validate(source, tokenCreator, null)
    .onComplete(testContext.asyncAssertFailure(e -> {
      assertThat(e, is(instanceOf(TokenValidationException.class)));
      assertThat(((TokenValidationException) e).getHttpResponseCode(), is(500));
    }));

For me assertThat instead of assertTrue is more readable, and gives better error messages if the assertion fails.

adamdickmeiss: It appears all validate implementation has no async stuff. But if they have (async stuff), only the latter will work.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,249 @@
...
+      return true;
+    }
+
+    return false;
+  }
+
+  /**
+   * Will return true if this this token requires a check that the user is active.
+   * @param requestHeaders The headers in the http context where the token is being provided.
+   * @return True if the check should be made.
```
adamdickmeiss, line 94: Please fix javadoc param

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,249 @@
...
+  /**
+   * Validates the provided token. Validation includes checking everything needed
+   * to determine whether the token should be authorized, including the signature and
+   * any special validation required by its type.
+   * @param sourceToken The JWT or JWE token to validate.
+   * @param tokenCreator The TokenCreator used to encode the token.
+   * @param request The request in the http context where the token is being provided.
+   * @return Future<Token> A Future containing the Token if it has passed validation.
+   * The Future may also contain a TokenValidationException if the validation has failed.
+   * @see TokenValidationException.
```
adamdickmeiss, line 55: Please remove the trailing period

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,249 @@
...
+        String tokenContent = tokenCreator.decodeJWEToken(sourceToken);
+        claims = new JsonObject(tokenContent);
+      } else {
+        tokenCreator.checkJWTToken(sourceToken);
+        claims = getClaims(sourceToken);
+      }
+    } catch (ParseException p) {
+      throw new TokenValidationException(invalidTokenMsg, p, 401);
+    } catch (JOSEException j) {
+      throw new TokenValidationException(invalidTokenMsg, j, 401);
```
adamdickmeiss, line 216: You can merge these two catch statements

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,249 @@
...
+   * @param tokenCreator The TokenCreator used to encode the token.
+   * @param request The request in the http context where the token is being provided.
+   * @return Future<Token> A Future containing the Token if it has passed validation.
+   * The Future may also contain a TokenValidationException if the validation has failed.
+   * @see TokenValidationException.
+   */
+  public static Future<Token> validate(String sourceToken,
+                                       TokenCreator tokenCreator,
+                                       HttpServerRequest request) {
+    Token token = null;
```
adamdickmeiss, line 60: Redundant assignment.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,249 @@
...
+    String encodedClaims = claims.encode();
+    return tokenCreator.createJWTToken(encodedClaims);
+  }
+
+  /**
+   * Encodes the token as a JWE token.
+   * @param tokenCreator The TokenCreator to use to encode the token.
+   * @return The encoded token.
+   * @throws JOSEException
+   * @throws ParseException
```
adamdickmeiss, line 118: Does not throw this

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -0,0 +1,249 @@
...
+    } catch (TokenValidationException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new TokenValidationException("Unexpected token validation exception", e, 500);
+    }
+  }
+
+  private static Token parse(String sourceToken, TokenCreator tokenCreator) throws TokenValidationException {
+    Token token = null;
+    JsonObject claims = null;
```
adamdickmeiss, line 202: Redundant assignment.

adamdickmeiss (approved):

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Done

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Done, thanks for catching.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Done thanks for noticing this.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: I'm not sure these should be combined. `checkJWToken` throws all three of these exceptions and I think it is desirable to pass these along to the `TokenValidationException` which allows for each to be logged in the module's error logger. But maybe I'm misunderstanding.

src/test/java/org/folio/auth/authtokenmodule/TokenTest.java
steveellis: Julian thanks for letting me know about the hamcrest `CoreMatchers`. I do agree that it makes the intent of the tests easier to scan and made a change based on that. In this case the access token will never have async operations in its validation so I think it is better to not use `asyncAssertFailure` here, although I'm glad to know about that and will put it in my toolbox.

julianladisch (approved):
Looks good, thanks!

PR #108, opened 2022-01-13
Clear permission cache when user logs in MODAT-118

adamdickmeiss:


steveellis (approved):
Seems fine since it is just removing the cache entry for the one user.

julianladisch (approved):
Looks good, thanks!

PR #109, opened 2022-01-13
Redundant web service request MODAT-117

adamdickmeiss:
When mod-authtoken fetches permissions for a user it calls

/perms/users?query=userId=={userId}
followed by:

/perms/users/{permUserId}/permissions?expanded=true
This can be changed to one call:

/perms/users/{userId}/permissions?expanded=true&indexField=userId


julianladisch (approved):
Looks good, thanks!

PR #110, opened 2022-01-24
Allow /token to create dummy tokens MODAT-119

adamdickmeiss:
Should fix "Unable to reset password on FOLIO Snapshot through UI".
This was possible before MODAT-109 and is used by mod-users-bl.

### DIFF
--- a/src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
+++ b/src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
@@ -392,9 +392,13 @@ public class MainVerticle extends AbstractVerticle {
         permissionsSource.clearCacheUser(userId, tenant);
       }
       String username = payload.getString("sub");
-      String token = new AccessToken(tenant, username, userId).encodeAsJWT(tokenCreator);
-
-      JsonObject responseObject = new JsonObject().put("token", token);
+      Token token;
+      if (payload.getBoolean("dummy", Boolean.FALSE)) {
+        token = new DummyToken(tenant, payload.getJsonArray("extra_permissions"));
+      } else {
+        token = new AccessToken(tenant, username, userId);
+      }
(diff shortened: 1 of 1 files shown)


steveellis (approved):
I worried about this but wasn't careful enough. Making all tokens just be a bag of properties made them easy to work with in a way, but slippery when trying to pin down what was going on with them. Here the incoming payload was re-used to generate the token (because why not, you've already got a JsonObject), but led me to totally miss that two token types.

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -392,9 +392,13 @@ private void handleSignToken(RoutingContext ctx) {
         permissionsSource.clearCacheUser(userId, tenant);
       }
       String username = payload.getString("sub");
-      String token = new AccessToken(tenant, username, userId).encodeAsJWT(tokenCreator);
-
-      JsonObject responseObject = new JsonObject().put("token", token);
+      Token token;
+      if (payload.getBoolean("dummy", Boolean.FALSE)) {
```
steveellis, line 396: Shouldn't this be `Boolean.TRUE`? Think you can also test for `type` to be `dummy`. Just to make things extra fun there are two types of `DummyToken`.

adamdickmeiss: mod-login does not supply dummy. The default ought to be false (not dummy). mod-users-bl OTOH provides dummy=true. In any case, it's a huge security risk to allow a dummy token to be created like that.

steveellis: Indeed you are right. Thanks for the links. I can see that password reset is sending a dummy token and gets an access token in return. Agree that using the type claim here would be wrong. Maybe adding a comment here would help a bit to explain what is going on.

steveellis (approved):

PR #111, opened 2022-01-26
Preserve "sub" for dummy token MODAT-119 part two

adamdickmeiss:
https://issues.folio.org/browse/MODAT-119

### DIFF
--- a/src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
+++ b/src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
@@ -185,7 +185,7 @@ public class AuthTokenTest {
     JsonObject payloadAccess = new JsonObject()
       .put("user_id", userUUID)
       .put("tenant", tenant)
-      .put("sub", "jones");
+      .put("sub", "joe");
 
     Response r;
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -655,8 +655,10 @@ public void test1(TestContext context) throws JOSEException, ParseException {
       .then()
       .statusCode(201).contentType("application/json").extract().body().asString();
     JsonObject tokenResponse = new JsonObject(response);
+    OkapiToken okapiToken = new OkapiToken(tokenResponse.getString("token"));
+    context.assertEquals(payloadDummy.getString("sub"), okapiToken.getUsernameWithoutValidation());
```
julianladisch, line 659: More concise:

```
String token = given()
  ...
  .statusCode(201).contentType("application/json")
  .extract().path("token");
  context.assertEquals(payloadDummy.getString("sub"), new OkapiToken(token).getUsernameWithoutValidation());
```

The same below.

adamdickmeiss: sure

PR #112, opened 2022-01-26
Define permissions auth.{sign,signrefresh}.token MODAT-120

adamdickmeiss:
https://issues.folio.org/browse/MODAT-120

This change has been tested with an actual setup with core modules etc.


julianladisch (approved):
Looks good, thanks!

PR #113, opened 2022-02-10
MODAT-110: Implement token persistent store

steveellis:
This still has some TODOs which I will take care of.

- [x] Merge in the handful of bugfixes from master
- [x] Deploy and test on VM
- [x] Implement the `getApiTokensForTenant` method
- [x] Resolve and remove TODOs

Would be great however to get some feedback.

The main thing to notice is that most of what was once in `MainVerticle.java` is now in `AuthorizeApi.java`, but most of that code hasn't changed. `AuthorizeApi` implements the `postInit` method to handle things like creating the tenant-level tables for each of our new classes: one for api token storage and the other for refresh token storage.

The `AuthorizeApi` gets wired up to the `MainVerticle` in the `RouterCreator` array there. The [order of adding items](https://github.com/folio-org/mod-authtoken/blob/bb426856114e8f1f0d7bd64236e89e01b559e926/src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java#L56) to the `RouterCreator` is [opposite to the example](https://github.com/folio-org/folio-vertx-lib/blob/8cb25ee0709804ea6e9a065decf963225aa8b2f7/example/src/main/java/org/folio/tlib/example/MainVerticle.java#L23). If my order isn't used, the filter of `handleAuthorize` takes precedence and doesn't let [the call ](https://github.com/folio-org/mod-authtoken/blob/bb426856114e8f1f0d7bd64236e89e01b559e926/src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java#L1199)to `/_/tenant` succeed.


julianladisch:
> Would be great however to get some feedback.

This looks good! I don't see any major issues.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,154 @@
...
+    String createTable = "CREATE TABLE IF NOT EXISTS " +
+        tableName(tenant, API_TOKEN_SUFFIX) +
+        "(id UUID PRIMARY key, token TEXT NOT NULL, " +
+        "is_revoked BOOLEAN NOT NULL, issued_at INT8 NOT NULL)";
+
+    log.info("Creating {} tables", TokenStore.class.getName());
+
+    return pool.withConnection(conn -> {
+      return conn.query(createTable).execute().mapEmpty();
+    });
```
julianladisch, line 60:
```suggestion
    return pool.query(createTable).execute().mapEmpty();
```

PgPool has shortcuts for a single query or a single preparedQuery.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1042,168 @@ private String getMagicPermission(String endpoint) {
...
+  @Test
+  public void testStoreSaveRefreshToken(TestContext context) {
+    var ts = new RefreshTokenStore(vertx, tenant);
+    Async async = context.async();
+    var rt = new RefreshToken(tenant, "jones", userUUID, "http://localhost:" + port);
+    ts.saveToken(rt).onComplete(context.asyncAssertSuccess(x -> {
+      ts.checkTokenNotRevoked(rt).onComplete(context.asyncAssertSuccess(y -> {
+        async.complete();
+      }));
+    }));
```
julianladisch, line 1054: context.async() + async.complete() is not needed when using context.asyncAssertSuccess().
Future has built-in error handling.

```
    ts.saveToken(rt)
    .compose(x -> ts.checkTokenNotRevoked(rt))
    .onComplete(context.asyncAssertSuccess());
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1042,168 @@ private String getMagicPermission(String endpoint) {
...
+    var apiToken = new ApiToken(tenant);
+    ts.saveToken(apiToken).onComplete(context.asyncAssertSuccess(a -> {
+      ts.checkTokenNotRevoked(apiToken).onComplete(context.asyncAssertSuccess(b -> {
+        ts.revokeToken(apiToken).onComplete(context.asyncAssertSuccess(c -> {
+          ts.checkTokenNotRevoked(apiToken).onComplete(context.asyncAssertFailure(d -> {
+            async.complete();
+          }));
+        }));
+      }));
+    }));
```
julianladisch, line 1104:
```
    ts.saveToken(apiToken)
    .compose(x -> ts.checkTokenNotRevoked(apiToken))
    .compose(x -> ts.revokeToken(apiToken))
    .onComplete(context.asyncAssertSuccess())
    .compose(x -> ts.checkTokenNotRevoked(apiToken))
    .onComplete(context.asyncAssertFailure());
```

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
steveellis: I think you are right. Will give this a try.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: This is great. Yes much better that way. Thanks for taking a look Julian!

pom.xml
```diff
@@ -53,7 +53,7 @@
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-stack-depchain</artifactId>
-        <version>4.1.4</version>
+        <version>4.2.1</version>
```
julianladisch, line 56:
```suggestion
        <version>4.2.4</version>
```

pom.xml
```diff
@@ -121,6 +126,12 @@
       <version>3.3.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>1.16.2</version>
```
julianladisch, line 132:
```suggestion
      <version>1.16.3</version>
```

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+        token = new AccessToken(tenant, username, userId);
+      }
+      JsonObject responseObject = new JsonObject().put("token", token.encodeAsJWT(tokenCreator));
+      endJson(ctx, 201, responseObject.encode());
+    } catch (Exception e) {
+      endText(ctx, 400, e);
+    }
+  }
+
+  private void handleAuthorize(RoutingContext ctx) {
```
julianladisch, line 382: I have difficulties to understand this method because it is very long, more than 300 lines. Can you move parts of it into sub-methods?

```
  someSubMethod();
  if (ctx.response().ended()) {
    return;
  }
```

julianladisch: This will also make it easier to test some error handling code to increase code coverage: https://sonarcloud.io/component_measures?id=org.folio%3Amod-authtoken&metric=new_coverage&pullRequest=113&selected=org.folio%3Amod-authtoken%3Asrc%2Fmain%2Fjava%2Forg%2Ffolio%2Fauth%2Fauthtokenmodule%2FAuthorizeApi.java

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+        perms.add(permName);
+      }
+    }
+  }
+
+  public String extractToken(String authorizationHeader) {
+    // Grab anything after 'Bearer' and whitespace
+    Pattern pattern = Pattern.compile("Bearer\\s+(.+)");
+    Matcher matcher = pattern.matcher(authorizationHeader);
+    if (matcher.find() && matcher.groupCount() > 0) {
```
julianladisch, line 699:
```suggestion
    if (matcher.find()) {
```
`groupCount()` is always 1 when `find()` is true because `(.+)` is a required match.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,175 @@
...
+   * @param apiToken The API token to revoke.
+   * @return A failed future if the revoke operation failed. Otherwise a succeeded future
+   * is returned.
+   */
+  public Future<Void> revokeToken(ApiToken apiToken) {
+    UUID tokenId = apiToken.getId();
+    log.info("Revoking API token {}", tokenId);
+
+    String update = "UPDATE " + tableName(tenant, API_TOKEN_SUFFIX) +
+        "SET is_revoked=$1 WHERE id=$2";
```
julianladisch, line 141: `is_revoked=TRUE` can avoid `where` variable: `.execute(Tuple.of(tokenId))`

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  public Future<Void> removeAll(SqlConnection conn, String suffix) {
+    log.info("Removing all tokens from storage");
+
+    String delete = "DELETE FROM " + tableName(tenant, suffix);
+
+    return conn.preparedQuery(delete).execute().mapEmpty();
+  }
+
+  protected Future<Row> getRow(SqlConnection conn, String select, Tuple where) {
```
julianladisch, line 50: All calls to this method can be changed to call `getRow(String select, Tuple where)` so that this method can call `pool.preparedQuery(select)`

julianladisch: Please check whether `SqlConnection` parameter of the other methods of TokenStore can also been removed.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+    Tuple where = Tuple.of(Boolean.TRUE, tokenId);
+
+    return conn.preparedQuery(update).execute(where).mapEmpty();
+  }
+
+  private Future<Void> revokeAllTokensForUser(SqlConnection conn, UUID userId) {
+    log.info("Revoking all refresh tokens for user id {}", userId);
+
+    String update = "UPDATE " + tableName(tenant, REFRESH_TOKEN_SUFFIX) +
+        "SET is_revoked=$1 WHERE user_id=$2";
```
julianladisch, line 197: Can you create a database index for user_id to speed up this query?

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+
+    String delete = "DELETE FROM " + tableName(tenant, REFRESH_TOKEN_SUFFIX) +
+      "WHERE expires_at<$1";
+    Tuple where = Tuple.of(now);
+
+    return pool.preparedQuery(delete).execute(where).mapEmpty();
+  }
+
+  public Future<Integer> countTokensStored(String tenant) {
+    String select = "SELECT * FROM " + tableName(tenant, REFRESH_TOKEN_SUFFIX);
```
julianladisch, line 170: `SELECT count(*)` avoids transferring all records from database to mod-authtoken.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+   * expire.
+   * @return A failed future should the deletion fail. A succeeded future if the
+   * deletion succeeds.
+   */
+  public Future<Void> cleanupExpiredTokens() {
+    long now = Instant.now().getEpochSecond();
+    log.info("Cleaning up tokens which are older than: {}", now);
+
+    String delete = "DELETE FROM " + tableName(tenant, REFRESH_TOKEN_SUFFIX) +
+      "WHERE expires_at<$1";
```
julianladisch, line 163: Can you create a database index for expires_at to speed up this query?

julianladisch, line 163: `DELETE FROM ... WHERE id IN (SELECT id FROM ... WHERE expires_at<$1 FOR UPDATE SKIP LOCKED)`

This skips rows that are locked and therefore doesn't wait until they get unlocked.
This locks rows to be deleted and therefore avoids rollbacks.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  private Future<Void> setTokenRedeemed(SqlConnection conn, RefreshToken refreshToken) {
+    UUID tokenId = refreshToken.getId();
+    String tenant = refreshToken.getTenant();
+
+    log.info("Setting refresh token id {} to redeemed", tokenId);
+
+    String update = "UPDATE " + tableName(tenant, REFRESH_TOKEN_SUFFIX) +
+        "SET is_redeemed=$1 WHERE id=$2";
```
julianladisch, line 187: `is_redeemed=TRUE`

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
```diff
@@ -0,0 +1,79 @@
...
+        log.error(msg, where.toString());
+        return Future.failedFuture("Token not found");
+      }
+      Row row = rows.iterator().next();
+      return Future.succeededFuture(row);
+    });
+  }
+
+  protected Future<RowSet<Row>> getRows(SqlConnection conn, String select, Tuple where) {
+    return conn.preparedQuery(select).execute(where).map(rows -> rows);
```
julianladisch, line 63: `map(rows -> rows)` doesn't do anything and should be removed. The same below.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,176 @@ private String getMagicPermission(String endpoint) {
...
+        var s5 = ts.saveToken(at5);
+        return CompositeFuture.all(s1, s2, s3, s4, s5)
+          .compose(y -> ts.getApiTokensForTenant(tenant))
+          .compose(y -> {
+            assertThat(y.size(), is(5));
+            return Future.succeededFuture();
+          });
+    }).onComplete(context.asyncAssertSuccess(z -> {
+      async.complete();
+    }));
```
julianladisch, line 1188:
```suggestion
        return CompositeFuture.all(s1, s2, s3, s4, s5);
      })
      .compose(x -> ts.getApiTokensForTenant(tenant))
      .onComplete(context.asyncAssertSuccess(tokens -> {
        assertThat(tokens.size(), is(5));
      }));
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,176 @@ private String getMagicPermission(String endpoint) {
...
+    var at4 = new ApiToken(tenant);
+    var at5 = new ApiToken(tenant);
+
+    var ts = new ApiTokenStore(vertx, tenant, tokenCreator);
+
+    Async async = context.async();
+
+    ts.removeAll()
+      .compose(x -> {
+        var s1 = ts.saveToken(at1);
```
julianladisch, line 1175: `var s1 = ts.saveToken(new ApiToken(tenant));`
avoiding the at1 variable.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,176 @@ private String getMagicPermission(String endpoint) {
...
+            return Future.succeededFuture();
+          })
+          .compose(y -> ts.cleanupExpiredTokens())
+          .compose(y -> ts.countTokensStored(tenant))
+          .compose(y -> {
+            assertThat(y, is(3));
+            return Future.succeededFuture();
+           });
+    }).onComplete(context.asyncAssertSuccess(z -> {
+      async.complete();
```
julianladisch, line 1156:
```suggestion
        return CompositeFuture.all(s1, s2, s3, s4, s5);
      })
      .compose(x -> ts.countTokensStored(tenant))
      .onSuccess(count -> assertThat(count, is(5)))
      .compose(x -> ts.cleanupExpiredTokens())
      .compose(x -> ts.countTokensStored(tenant))
      .onSuccess(count -> assertThat(count, is(3)))
      .onComplete(context.asyncAssertSuccess());
```

src/test/java/org/folio/auth/authtokenmodule/MainVerticleTest.java
```diff
@@ -1,21 +1,21 @@
...
-import com.nimbusds.jose.JOSEException;
-
 import io.vertx.core.Vertx;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
 
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.MatcherAssert.assertThat;
+
```
julianladisch, line 16: Static imports usually go before non-static imports:
https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Constructs the refresh token store.
+   * @param vertx A reference to the current vertx object.
+   * @param tenant The tenant which is in scope for a given token. This can be obtained
+   * from the refresh token when the refresh token is created or arrives.
+   */
+  public RefreshTokenStore(Vertx vertx, String tenant) {
+    super(vertx, tenant);
+
+    // TODO implement this to clean up expired tokens from the store.
```
julianladisch, line 36: This implementation is missing.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+    // TODO Consider using an okapi timer.
+    // vertx.setPeriodic(delay, handler)
```
julianladisch, line 38: I rather prefer this approach:
After a new token has been inserted asynchronously delete expired tokens. No need to run the delete method when there is no activity.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Creates the table for this token store if it doesn't yet exist.
+   * @return A failed future should the save operation fail. Otherwise a succeeded future
+   * is returned, even if the table exists.
+   */
+  public Future<Void> createIfNotExists() {
```
julianladisch, line 46: `createTable` better says what it creates.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,202 @@
...
+    //    other requests could have revoked the token anytime. We can't let these through.
+    // 3. If isRedeemed is false, call setTokenRedeemed and return a success. It is not
+    //    revoked or redeemed. This is the desired state. RTs can be used only once.
+    //    Any subsequent uses must be considered as a leaked (compromised) token.
+    // 4. If isRedeemed is true call revokeAllTokensForUser and return a failed future.
+    //    isRedeemed true means "someone tried to use it a second time". This is the
+    //    second use of the token and it means the token has been leaked. We
+    //    consider the user's account compromised and revoke all their refresh tokens.
+    //    They will need to login again to get a new RefreshToken that isn't revoked
+    //    or redeemed.
```
julianladisch, line 103: This comment should be merged into the javadoc.

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  protected Future<RowSet<Row>> getRows(SqlConnection conn, String select) {
+    return conn.preparedQuery(select).execute().map(rows -> rows);
```
julianladisch, line 67: `map(rows -> rows)` does nothing and can be deleted.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,175 @@
...
+import io.vertx.sqlclient.Tuple;
+
+/**
+ * The API token store provides persistence for API tokens. API tokens need to be
+ * persisted so that they can be revoked.
+ */
+public class ApiTokenStore extends TokenStore {
+  private static final Logger log = LogManager.getLogger(ApiTokenStore.class);
+
+  private static String API_TOKEN_SUFFIX = "api_tokens";
```
julianladisch, line 27: final

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
steveellis: > All calls to this method can be changed to call `getRow(String select, Tuple where)` so that this method can call `pool.preparedQuery(select)`

I think with `getRow` I want to keep passing in the `SqlConnection`, because `RefreshTokenStore.checkTokenNotRevoked` uses it and should share the same connection for its operations so using `withConnection` makes good sense there.

But I think you are right that the other methods in `TokenStore`. They don't need this parameter and can use the pool directly.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Thank you. This is what I was going for but didn't quite get there.

steveellis:
My commit says "should resolve all of Julian's feedback". Should have said "most" since I left a few things open to discussion.

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
julianladisch: How about
```
  protected Future<Row> getRow(String select, Tuple where) {
    return pool.withConnection(conn -> getRow(conn, select, where));
  }
```

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
```diff
@@ -0,0 +1,60 @@
...
+
+  protected Future<Void> removeAll(String suffix) {
+    log.info("Removing all tokens from storage");
+
+    String delete = "DELETE FROM " + tableName(tenant, suffix);
+
+    return pool.preparedQuery(delete).execute().mapEmpty();
+  }
+
+  protected Future<Row> getRow(SqlConnection conn, String select, Tuple where) {
```
julianladisch, line 41: The SQL query `SELECT is_redeemed, is_revoked FROM table WHERE id=$1` consists of
* the SELECT clause `SELECT is_redeemed, is_revoked`
* the FROM clause `FROM table`
* the WHERE clause `WHERE id=$1`

Tuple.of(id) is the parameter tuple.

Therefore I don't expect the method parameters `String select, Tuple where` but `String sql, Tuple tuple` or `String sql, Tuple parameters`.

julianladisch, line 41: javadoc should mention that the first row of a multi-row result is returned and all other rows are fetched but ignored, and that the caller may append "LIMIT 1" to speed up the query.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,187 @@
...
+
+    log.info("Inserting token id {} into {} token store", id, API_TOKEN_SUFFIX);
+
+    String insert = "INSERT INTO " + tableName(tenant, API_TOKEN_SUFFIX) +
+        "(id, token, is_revoked, issued_at) VALUES ($1, $2, $3, $4)";
+    var values = Tuple.of(id, token, isRevoked, issuedAt);
+
+    return pool.preparedQuery(insert).execute(values).mapEmpty();
+
+    // TODO Should we return the encoded API token to callers?
```
julianladisch, line 95: yes

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,187 @@
...
+
+    String select = "SELECT is_revoked FROM " + tableName(tenant, API_TOKEN_SUFFIX) +
+      "WHERE id=$1";
+    Tuple where = Tuple.of(tokenId);
+
+    return pool.withConnection(conn -> {
+      return getRow(conn, select, where).compose(row -> {
+        Boolean isRevoked = row.getBoolean("is_revoked");
+
+        log.info("Revoked status of {} token id {} is {}", API_TOKEN_SUFFIX, tokenId, isRevoked);
```
julianladisch, line 119: Log messages slow down the software. Can you go through your pull request and reduce the log level from info to debug for log messages that are not needed for production? No info message is needed at all if the token is not revoked because other log messages show that the request to mod-authtoken is successful.

src/main/java/org/folio/auth/authtokenmodule/MissingAlgorithmException.java
```diff
@@ -0,0 +1,7 @@
+package org.folio.auth.authtokenmodule;
+
+public class MissingAlgorithmException extends Exception {
```
julianladisch, line 3: Can `Exception` been replaced by `RuntimeException` or `Error` to make it an unchecked exception because we cannot recover from it?

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,216 @@
...
+        tableName(tenant, REFRESH_TOKEN_SUFFIX) + "(expires_at)";
+    String createIndexUserId = "CREATE INDEX IF NOT EXIST ON " +
+        tableName(tenant, REFRESH_TOKEN_SUFFIX) + "(user_id)";
+
+    log.info("Creating {} tables", RefreshTokenStore.class.getName());
+
+    return pool.query(createIndexExpiresAt).execute()
+      .compose(x -> pool.query(createIndexUserId).execute())
+      .mapEmpty();
+  }
```
julianladisch, line 71: createIndexesIfNotExists() can be merged into createTableIfNotExists(). It doesn't make sense to create the table without indexes.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,216 @@
...
+   * is returned.
+   */
+  public Future<Void> checkTokenNotRevoked(RefreshToken refreshToken) {
+    UUID tokenId = refreshToken.getId();
+    UUID userId = refreshToken.getUserId();
+    String table = tableName(tenant, REFRESH_TOKEN_SUFFIX);
+
+    log.info("Checking token redeemed of token id {} for tenant {}", tokenId, tenant);
+
+    String select = "SELECT is_redeemed, is_revoked FROM " + table + "WHERE id=$1";
```
julianladisch, line 126: Sending two request with the same refresh token in parallel to mod-authtoken may successfully create two new refresh tokens because there is a race condition in the time interval between the SELECT and the UPDATE statement.

To avoid the race condition they can be combined into an atomic operation:

```
UPDATE table SET is_redeemed=TRUE WHERE id=$1 AND is_redeemed=FALSE and is_revoked=FALSE
```

The returned rowCount is either 1 or 0.
1 indicates success.
0 indicates that the refresh token doesn't exists, or that is_redeemed or is_revoked is true. This is a failure. Simply SELECT the record and take the needed actions.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,216 @@
...
+   *    second use of the token and it means the token has been leaked. We
+   *    consider the user's account compromised and revoke all their refresh tokens.
+   *    They will need to login again to get a new RefreshToken that isn't revoked
+   *    or redeemed.
+   *
+   * @param refreshToken The RefreshToken to check.
+   * @return A failed future if the token has been revoked. Otherwise a succeeded future
+   * is returned.
+   */
+  public Future<Void> checkTokenNotRevoked(RefreshToken refreshToken) {
```
julianladisch, line 119: This method not only checks, it actually redeems the token. A better method name might be `redeemToken`.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  /**
+   * Gets a list of all the API tokens for a given tenant.
+   * @param tenant The tenant.
+   * @return A list of all the API tokens for a tenant.
+   */
+  public Future<List<ApiToken>> getApiTokensForTenant(String tenant) {
+    String select = "SELECT token FROM " + tableName(tenant, API_TOKEN_SUFFIX);
+    List<ApiToken> tokens = new ArrayList<ApiToken>();
```
julianladisch, line 152:
```suggestion
    List<ApiToken> tokens = new ArrayList<>();
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,161 @@ private String getMagicPermission(String endpoint) {
...
+    ts.saveToken(apiToken).compose(x ->  ts.checkTokenNotRevoked(apiToken))
+      .onComplete(context.asyncAssertSuccess());
+  }
+
+  @Test
+  public void testStoreApiTokenNotFound(TestContext context) {
+    var ts = new ApiTokenStore(vertx, tenant, tokenCreator);
+    // A ApiToken which doesn't exist in storage is treated as revoked.
+    var unsavedToken = new ApiToken(tenant);
+    ts.checkTokenNotRevoked(unsavedToken).onComplete(context.asyncAssertFailure());
```
julianladisch, line 1076: Can you test the error message?
```
.onComplete(context.asyncAssertFailure(e -> {
  assertThat(e.getMessage(), is(stringContaining("revoked")));
}));
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,161 @@ private String getMagicPermission(String endpoint) {
...
+    var ts = new ApiTokenStore(vertx, tenant, tokenCreator);
+    var apiToken = new ApiToken(tenant);
+    ts.saveToken(apiToken).compose(x ->  ts.checkTokenNotRevoked(apiToken))
+      .onComplete(context.asyncAssertSuccess());
+  }
+
+  @Test
+  public void testStoreApiTokenNotFound(TestContext context) {
+    var ts = new ApiTokenStore(vertx, tenant, tokenCreator);
+    // A ApiToken which doesn't exist in storage is treated as revoked.
```
julianladisch, line 1074: This comment is missing from checkTokenNotRevoked's javadoc.
The error message should reflect that: "The token doesn't exist or has been revoked."

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,161 @@ private String getMagicPermission(String endpoint) {
...
+      .compose(a -> ts.checkTokenNotRevoked(rt2))
+      .onComplete(context.asyncAssertSuccess()) // First check should succeed.
+      .compose(a -> ts.checkTokenNotRevoked(rt2))
+      .onComplete(context.asyncAssertFailure(b ->  { // Second should fail.
+        // At this point all tokens for the user should be revoked so any subsequent
+        // checks should fail.
+        ts.checkTokenNotRevoked(rt1).onComplete(context.asyncAssertFailure(c -> {
+          ts.checkTokenNotRevoked(rt3).onComplete(context.asyncAssertFailure(d -> {
+            async.complete();
+          }));
```
julianladisch, line 1113:
```
          context.asyncAssertFailure(d -> {
            async.complete();
          })
```
can always be replaced by
```
          context.asyncAssertFailure()
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,161 @@ private String getMagicPermission(String endpoint) {
...
+        var s5 = ts.saveToken(rt5);
+        return CompositeFuture.all(s1, s2, s3, s4, s5);
+      })
+      .compose(y -> ts.countTokensStored(tenant))
+      .onSuccess(y -> assertThat(y, is(5)))
+      .compose(y -> ts.cleanupExpiredTokens())
+      .compose(y -> ts.countTokensStored(tenant))
+      .onSuccess(y -> assertThat(y, is(3)))
+      .onComplete(context.asyncAssertSuccess(y -> {
+        async.complete();
```
julianladisch, line 1151: async.complete() is not needed inside of context.asyncAssertSuccess

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,161 @@ private String getMagicPermission(String endpoint) {
...
+        var s2 = ts.saveToken(new ApiToken(tenant));
+        var s3 = ts.saveToken(new ApiToken(tenant));
+        var s4 = ts.saveToken(new ApiToken(tenant));
+        var s5 = ts.saveToken(new ApiToken(tenant));
+        return CompositeFuture.all(s1, s2, s3, s4, s5);
+      })
+      .compose(y -> ts.getApiTokensForTenant(tenant))
+      .onSuccess(y -> assertThat(y.size(), is(5)))
+      .onComplete(context.asyncAssertSuccess(z -> {
+        async.complete();
```
julianladisch, line 1172: async.complete() is not needed inside of context.asyncAssertSuccess

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,161 @@ private String getMagicPermission(String endpoint) {
...
+        var s1 = ts.saveToken(new ApiToken(tenant));
+        var s2 = ts.saveToken(new ApiToken(tenant));
+        var s3 = ts.saveToken(new ApiToken(tenant));
+        var s4 = ts.saveToken(new ApiToken(tenant));
+        var s5 = ts.saveToken(new ApiToken(tenant));
+        return CompositeFuture.all(s1, s2, s3, s4, s5);
+      })
+      .compose(y -> ts.getApiTokensForTenant(tenant))
+      .onSuccess(y -> assertThat(y.size(), is(5)))
+      .onComplete(context.asyncAssertSuccess(z -> {
```
julianladisch, line 1171: onSuccess and onComplete(context.asyncAssertSuccess can be combined:
```
      .onComplete(context.asyncAssertSuccess(tokens -> assertThat(tokens.size(), is(5))));
```

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
julianladisch: Can you add a check for expires_at to reject the token if it is expired?

src/main/java/org/folio/auth/authtokenmodule/tokens/RefreshToken.java
```diff
@@ -10,12 +9,33 @@
...
+    return UUID.fromString(claims.getString("user_id"));
+  }
+
+  public long getExpiresAt() {
+    return claims.getLong("exp");
+  }
+
+  /**
+   * Should only be used by tests.
+   * @param to The epoch seconds time stamp to the iat claim to.
```
julianladisch, line 33:
```suggestion
   * @param to The epoch seconds time stamp to set the iat claim to.
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -53,6 +69,9 @@
   static Vertx vertx;
   Async async;
 
+  @ClassRule
+  public static PostgreSQLContainer<?> postgresSQLContainer = TokenStoreTestContainer.create();;
```
julianladisch, line 73:
```suggestion
  public static PostgreSQLContainer<?> postgresSQLContainer = TokenStoreTestContainer.create();
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1023,4 +1043,161 @@ private String getMagicPermission(String endpoint) {
...
+    var rt = new RefreshToken(tenant, "jones", userUUID, "http://localhost:" + port);
+    ts.saveToken(rt)
+      .compose(x -> ts.checkTokenNotRevoked(rt))
+      .onComplete(context.asyncAssertSuccess());
+  }
+
+  @Test
+  public void testStoreRefreshTokenNotFound(TestContext context) {
+    var ts = new RefreshTokenStore(vertx, tenant);
+    // A RefreshToken which doesn't exist is treated as revoked.
```
julianladisch, line 1058: This comment should also been added to the checkTokenNotRevoked method.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
steveellis: Agree. Do you have an opinion here on what _should_ be useful info messages? I'm thinking perhaps:
* When a RT token has been detected as leaked and revoked
* When an ApiToken has been revoked

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,187 @@
...
+  //   List<String> tokens = new ArrayList<String>();
+  //   return pool.query(select).execute().compose(rows -> {
+  //     for (Row row : rows) {
+  //       String tokenString = row.getString("token");
+  //       tokens.add(tokenString);
+  //     }
+  //     log.info("Retrieved {} token rows for tenant {}", rows.rowCount(), tenant);
+  //     return Future.succeededFuture(tokens);
+  //   });
+  // }
```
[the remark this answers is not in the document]
steveellis: This is just me being sloppy. What I was going to do was replace the method with the commented method so that clients get string representations of the tokens rather than token objects. I think clients will want strings and not care at all about objects since the primary purpose of this method will be for clients to get access to the tokens for use in other contexts and users will basically copy paste them.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: I'm finding that without `async.complete()` these test methods don't complete and time out.

src/main/java/org/folio/auth/authtokenmodule/tokens/RefreshToken.java
steveellis: Thanks. It's also to `set the exp claim` rather than `iat`.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
steveellis: This is a great idea and also a nice simplification since there's no longer a need to select anything either before or after.

> Can you add a check for expires_at to reject the token if it is expired?

Added. I opted to just check the provided rather than the stored token since the provided token shouldn't have been tampered with. See what you think.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
steveellis: I'm no longer seeing a reason for the concept of token redemption. I think semantically the concept exists since it has two "modes" but logically it really only needs to have one. It's either revoked or not. When it is created it's not revoked and can be redeemed, and once it is provided for redemption it's now no longer "good" -- it's used and can't be used again, and therefore revoked.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
steveellis: I broke this out into a separate method because I was having trouble having subsequent database operations complete after I created the indexes. In other words, I couldn't create the RT table, create its indexes, _and then_ create the API token table. Instead that third operation would silently fail. I didn't resolve this except to make the index creation the last thing done. Would love to know what might be going on there.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,216 @@
...
+        "SET is_redeemed=TRUE WHERE id=$1";
+
+    return conn.preparedQuery(update).execute(Tuple.of(tokenId)).mapEmpty();
+  }
+
+  private Future<Void> revokeAllTokensForUser(SqlConnection conn, UUID userId) {
+    log.info("Revoking all refresh tokens for user id {}", userId);
+
+    String update = "UPDATE " + tableName(tenant, REFRESH_TOKEN_SUFFIX) +
+        "SET is_revoked=$1 WHERE user_id=$2";
```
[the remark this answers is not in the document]
steveellis: Implemented. Or at least I think I did. From what I'm gathering here is that both operations should be run right after the other. The first is being "aggressive" to get everything that isn't locked, and the second will block until any locked rows are released.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Yes will add in error message checking.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Added the error message to `getRow` and also now check for it in the test.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: See my comment above. Sometimes I need it. Sometimes I don't. Any idea what is going on?

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Done

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Noticing that it is only when I do `CompositeFuture.all()` that I need `async.complete()`.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,199 @@
...
+    log.info("Creating {} tables", RefreshTokenStore.class.getName());
+
+    // Refresh tokens have an owning user, but the token itself isn't persisted
+    // since it isn't used for anything. Just the id is enough.
+    String createTable = "CREATE TABLE IF NOT EXISTS " +
+        tableName(tenant, REFRESH_TOKEN_SUFFIX) +
+        "(id UUID PRIMARY key, user_id UUID NOT NULL, " +
+        "is_revoked BOOLEAN NOT NULL, expires_at INT8 NOT NULL)";
+
+    String createIndexExpiresAt = "CREATE INDEX IF NOT EXIST ON " +
```
julianladisch, line 52:
```suggestion
    String createIndexExpiresAt = "CREATE INDEX IF NOT EXISTS expires_at_idx ON " +
```
The same for the user_id index.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,199 @@
...
+    return conn.preparedQuery(updateSkipLocked).execute(params)
+      .compose(x -> conn.preparedQuery(updateAll).execute(params)).mapEmpty();
+  }
+
+  /**
+   * Cleans up (deletes) refresh tokens which have passed their time of expiration.
+   * Clients need to ensure that they are requesting new refresh tokens before they
+   * expire.
+   * @return A failed future should the deletion fail. A succeeded future if the
+   * deletion succeeds.
```
julianladisch, line 175: This comment is not needed. This behavior is the default for any `Future<Void>`.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1016,11 +1039,187 @@ public void testAdminHealth()  {
...
+        assertThat(b.getMessage(), containsString("revoked"));
+      }))
+      .compose(a -> ts.checkTokenNotRevoked(rt1))
+      .onComplete(context.asyncAssertFailure(b -> {
+        assertThat(b.getMessage(), containsString("revoked"));
+      }))
+      .compose(a -> ts.checkTokenNotRevoked(rt3))
+      .onComplete(context.asyncAssertFailure(b -> {
+        assertThat(b.getMessage(), containsString("revoked"));
+        async.complete();
```
julianladisch, line 1137: This async is not needed because context.asyncAssertFailure() has is included.

steveellis: See my comment above. Without `async.complete` I get a timeout. I believe `async.complete` is needed when you have a `CompositeFuture` in the mix,since only those methods which use a composite future have this characteristic. Give the code a try to see what I mean.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
julianladisch: context.asyncAssertSuccess(x -> {
        // some tests without callback
        async.complete();
      }

This async is not needed because there is a context.asyncAssertSuccess that already contains an async, and there is no other lambda.

An async is only needed if there is no asyncAssertXXX:

    ts.checkTokenNotRevoked(rt1)
    .onComplete(a -> async.complete());

or

    ts.checkTokenNotRevoked(rt1)
    .compose(a -> ts.checkTokenNotRevoked(rt2)
                    .onComplete(x -> async.complete()))

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
steveellis: Removed

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
steveellis: Fixed

steveellis:
Was able to test out the latest on the VM today and it was fine.

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class TokenStore {
+  private static final Logger log = LogManager.getLogger(TokenStore.class);
+
+  protected Vertx vertx;
+  protected String tenant;
+  protected TenantPgPool pool;
+
+  public TokenStore(Vertx vertx, String tenant) {
+    this.vertx = vertx;
```
adamdickmeiss, line 26: No need for vertx handle member in this class

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  /**
+   * Gets the fully qualified table name for the tenant.
+   * @param tenant The tenant string
+   * @param tableNameSuffix The part of the table name after the dot.
+   * @return Returns the table name with a space appended to the end of the string.
+   */
+  protected String tableName(String tenant, String tableNameSuffix) {
+    return pool.getSchema() + "." + tableNameSuffix + " ";
```
adamdickmeiss, line 69: Parameter tenant not used. But even more important if it was used. It's a member of the class. So supplying anything but the same tenant is a programming error.

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+  }
+
+  private JsonObject parseEncryptionRequest(RoutingContext ctx, String[] requiredProperties) {
+    if (ctx.request().method() != HttpMethod.POST) {
+      String message = "Invalid method for this endpoint";
+      endText(ctx, 400, message);
+      return null;
+    }
+    String content = ctx.getBodyAsString();
+    JsonObject requestJson = null;
```
adamdickmeiss, line 199: No need to assign.

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+        String username = token.getClaims().getString("sub");
+        String userId = token.getClaims().getString("user_id");
+        String tenant = token.getClaims().getString("tenant");
+
+        try {
+          // TODO To do RTR we need to return both a new AT and a new RT here.
+          String at = new AccessToken(tenant, username, userId).encodeAsJWT(tokenCreator);
+          JsonObject responseObject = new JsonObject().put("token", at);
+          endJson(ctx, 201, responseObject.encode());
+          return;
```
adamdickmeiss, line 263: No need for this return

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+        } catch (Exception e) {
+          endText(ctx, 500, String.format("Unanticipated exception creating access token: %s", e.getMessage()));
+          return;
```
adamdickmeiss, line 266: No need for this return

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+      String msg, String unexpectedExceptionMsg) {
+    if (h instanceof TokenValidationException) {
+      var e = (TokenValidationException) h;
+      logger.error("{}: {}", msg, e.toString());
+      endText(ctx, e.getHttpResponseCode(), msg);
+      return;
+    }
+    logger.error("{}: {}", unexpectedExceptionMsg, h.toString());
+    endText(ctx, 500, unexpectedExceptionMsg);
+    return;
```
adamdickmeiss, line 284: No need for this return

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+      Future<Boolean> activeUser = Future.succeededFuture(Boolean.TRUE);
+      if (token.shouldCheckIfUserIsActive(finalUserId)) {
+        activeUser = userService.isActiveUser(finalUserId, tenant, okapiUrl, userRequestToken, requestId);
+      }
+      Future<PermissionData> retrievedPermissionsFuture = activeUser.compose(b -> {
+        if (TRUE.equals(b)) {
+          return permService.expandSystemPermissions(extraPermissions, tenant, okapiUrl, permissionsRequestToken,
+              requestId);
+        } else {
+          String msg = String.format("Invalid token: user with id {} is not active", finalUserId);
```
adamdickmeiss, line 580: Bad format. Shouldn't that be `%s`?

src/main/java/org/folio/auth/authtokenmodule/AuthorizeApi.java
```diff
@@ -0,0 +1,724 @@
...
+    Router router = Router.router(vertx);
+    router.route("/*").handler(BodyHandler.create());
+    router.route("/*").handler(this::handleAuthorize);
+    return Future.succeededFuture(router);
+  }
+
+  @Override
+  public Future<Void> postInit(Vertx vertx, String tenant, JsonObject tenantAttributes) {
+    logger.info ("Calling postInit for {}", AuthorizeApi.class.getName());
+
```
adamdickmeiss, line 135:
```suggestion
    if (!tenantAttributes.containsKey("module_to")) {
      return Future.succeededFuture(); // doing nothing for disable
    }
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -106,6 +127,8 @@ public static void setUpClass(TestContext context) throws NoSuchAlgorithmExcepti
         });
       }
     });
+
+    initializeTenantForTokenStore(tenant, new JsonObject());
```
adamdickmeiss, line 131:
```suggestion
    initializeTenantForTokenStore(tenant, new JsonObject().put("module_to", "mod-authtoken-1.0.0"));
```

src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java
```diff
@@ -1,754 +1,82 @@
...
-    }
-    setLogLevel(Level.toLevel(name));
-  }
-
-  static Level setLogLevel(Level level) {
-    Level existing = LogManager.getRootLogger().getLevel();
-    Configurator.setAllLevels(LogManager.getRootLogger().getName(), level);
-    return existing;
-  }
+  private TokenCreator tokenCreator;
```
adamdickmeiss, line 29: This can be a local variable.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,158 @@
...
+        "(id, token, is_revoked, issued_at) VALUES ($1, $2, $3, $4)";
+    Tuple params = Tuple.of(id, token, isRevoked, issuedAt);
+
+    return pool.preparedQuery(sql).execute(params).map(token);
+  }
+
+  /**
+   * Check that the token has not been revoked. This will return a failed future if
+   * the token has been revoked, otherwise it will return a succeeded future.
+   * @param refreshToken The API token to check.
```
adamdickmeiss, line 95: refreshToken not a parameter.

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * The API token store provides persistence for API tokens. API tokens need to be
+ * persisted so that they can be revoked.
+ */
+public class ApiTokenStore extends TokenStore {
+  private static final Logger log = LogManager.getLogger(ApiTokenStore.class);
+
+  private static final String API_TOKEN_SUFFIX = "api_tokens";
+  private TokenCreator tokenCreator;
```
adamdickmeiss, line 26: Can be final

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,158 @@
...
+   * @return A failed future should the save operation fail. Otherwise a succeeded future
+   * is returned. If the future succeeds, a string representation of the token is returned.
+   */
+  public Future<String> saveToken(ApiToken apiToken) {
+    UUID id = apiToken.getId();
+    long issuedAt = apiToken.getIssuedAt();
+    boolean isRevoked = false;
+    String tenant = apiToken.getTenant();
+
+    String token = "";
```
adamdickmeiss, line 75: No need to initialize

src/main/java/org/folio/auth/authtokenmodule/storage/ApiTokenStore.java
```diff
@@ -0,0 +1,158 @@
...
+  private TokenCreator tokenCreator;
+
+  /**
+   * Constructs the store for the given tenant. The tenant can be obtained from
+   * the token when it arrives for validation or when it is created.
+   * @param vertx A reference to the current vertx object.
+   * @param tenant The tenant which is in scope for a given token. The tenant
+   * can be obtained from the token when it arrives for validation or when it is
+   * created.
+   * @param tokenCreator A reference to the the current TokenCreator object
```
adamdickmeiss, line 35: Double the

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        // The row count is not 1 so the token has been used more than once or has been
+        // revoked when another token has been used more than once.
+        String leakedMessage = "Refresh token {} is revoked." +
+          " Revoking all tokens for user {}.";
+        log.info(leakedMessage, tokenId, userId);
+
+        return revokeAllTokensForUser(conn, userId).compose(y -> {
+          return Future.failedFuture("Token leaked. All tokens for user are now revoked.");
```
adamdickmeiss, line 139:
```suggestion
        return revokeAllTokensForUser(conn, userId)
          .compose(y -> Future.failedFuture("Token leaked. All tokens for user are now revoked."));
```

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  private boolean tokenHasExpired(RefreshToken rt) {
+    return Instant.now().getEpochSecond() >= rt.getExpiresAt();
+  }
+
+  private Future<Void> revokeAllTokensForUser(SqlConnection conn, UUID userId) {
+    log.debug("Revoking all refresh tokens for user id {}", userId);
+
+    // First attempt to update but skipping any locks to be the most aggressve as we can.
```
adamdickmeiss, line 152:
```suggestion
    // First attempt to update but skipping any locks to be the most aggressive as we can.
```

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
adamdickmeiss: Test enabling of the module.

adamdickmeiss (changes requested):
The `_tenant` interface is not declared in descriptors/ModuleDescriptor-template.json, so Okapi will never call the tenant API.

adamdickmeiss:
I can't my comment on why I requested changes, so I'll repeat it to be sure.. There's no `_tenant` interface in the module descriptor for this module, so the storage will not be initialized for any tenant. Just add a section as seen here: https://github.com/folio-org/mod-eusage-reports/blob/5c9cd0b68327d800d373bb067adf3e3776861c1a/descriptors/ModuleDescriptor-template.json#L6

I did that, locally here, and ran with my local build .. That failed. There's something wrong with the filtering and the tenant API call - something you also mention. I'll look into that soon.

descriptors/ModuleDescriptor-template.json
```diff
@@ -20,6 +20,14 @@
...
           "permissionsRequired": []
+        },
+        {
+          "methods": [ "POST" ],
+          "pathPattern": "/_/tenant"
+        },
+        {
+          "methods": [ "GET", "DELETE" ],
+          "pathPattern": "/_/tenant/{id}"
         }
```
adamdickmeiss, line 31: These endpoints *must* be part of the _tenant version 2 interface. So authtoken provides two interfaces: `authtoken` and `_tenant.`

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1016,11 +1038,187 @@ public void testAdminHealth()  {
...
+        var s1 = ts.saveToken(rt1);
+        var s2 = ts.saveToken(rt2);
+        var s3 = ts.saveToken(rt3);
+        var s4 = ts.saveToken(rt4);
+        var s5 = ts.saveToken(rt5);
+        return CompositeFuture.all(s1, s2, s3, s4, s5);
+      })
+      .compose(y -> ts.countTokensStored(tenant))
+      .onComplete(context.asyncAssertSuccess(count -> {
+        assertThat(count, is(3));
```
steveellis, line 1171: The problem with this is that I call `cleanupExpiredTokens` in `onComplete` in `saveToken` to get the "fire and forget" behavior that we want. Because of this there is no guarantee here that `cleanupExpiredTokens` will complete in the test. To keep things simple I could test `cleanupExpiredTokens` separately (which I originally had) but I'll also need a `saveToken` method which doesn't clean up to do that. Is there a better way?

src/main/java/org/folio/auth/authtokenmodule/storage/TokenStore.java
steveellis: Thanks for catching this. I think at one point I was not using `getSchema` and neglected to remove it.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
adamdickmeiss:
```suggestion
  public Future<Void> saveToken(RefreshToken refreshToken) {
    return saveToken(refreshToken, Promise.promise());
  }
  
  Future<Void> saveToken(RefreshToken refreshToken, Promise<Void> promise) {
   ..
   
    return pool.preparedQuery(insert).execute(values)
      .onComplete(x -> cleanupExpiredTokens().onComplete(promise)).mapEmpty();
  }
  
```

julianladisch: I'd use https://github.com/awaitility/awaitility

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
julianladisch: A gave it a try before and I still get no timeout locally. Jenkins also has no timeout: https://github.com/folio-org/mod-authtoken/pull/116

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
julianladisch: A gave it a try before and I still get no timeout locally.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Thank you for your branch Julian! I was having a hard time seeing what you wanted me to see. I had to stare at it for a while before I realized what was going on. I was only commenting out `async.complete()` and testing that. But I also had `Async async = context.async()`. Having `context.async()` alone, without the completion was the timeout's cause.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Thanks for the ideas. Julian that looks like a cool library but I wasn't able to figure out how to get it to play nice with vertx futures in the time I gave myself. I went with something like what Adam has but used a boolean instead to control which method of waiting gets called. See what you think.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -62,12 +62,15 @@ public RefreshTokenStore(Vertx vertx, String tenant) {
 
   /**
    * Save the token to the token store. This should be done anytime a new refresh token
-   * is issued.
+   * is issued. Also cleans up any expired tokens when called.
    * @param refreshToken The refresh token to store.
+   * @param waitForCleanupExpired True if the caller should wait until the cleanup of expired
+   * tokens succeeds, otherwise return after the save is complete and clean up the expired
+   * tokens in the background.
```
julianladisch, line 80: True if the returned Future completes when the cleanup completes, false if it completes when the save completes.

src/main/java/org/folio/auth/authtokenmodule/storage/RefreshTokenStore.java
```diff
@@ -62,12 +62,15 @@ public RefreshTokenStore(Vertx vertx, String tenant) {
...
    * @return A failed future should the save operation fail. Otherwise a succeeded future
    * is returned.
```
julianladisch, line 82: The completion of the save, or the completion of both the save and the cleanup if waitForCleanupExpired is true.

julianladisch (approved):
Looks good, thanks!

adamdickmeiss (changes requested):
Method postInit always creates tables. It should only do so on enable (not disable).

steveellis:
> Method postInit always creates tables. It should only do so on enable (not disable).

I think all I needed to do was add a check for `module_to` and return. Let me know if I have that right.

steveellis:
But I neglected to the property to my tenant  invoke. Let me fix that.

adamdickmeiss (approved):

PR #118, opened 2022-02-22
Always invoke filter (except for admin health).

adamdickmeiss:
The tenant call when recognized is just postponed to be handled later.


julianladisch (approved):
Looks good, thanks!

PR #120, opened 2022-02-25
MODAT-123: Report "no token" when permission required and token missing

julianladisch:
In addition report the user and userId when rejecting for unsufficient permissions.


steveellis (approved):
Looks good!

adamdickmeiss (approved):
Thx

PR #122, opened 2022-03-16
MODAT-126: Implement OpenApi

steveellis:
Implements OpenApi for all non-filter routes. Removed code for things that OpenAPi takes care of like HTTP method validation and schema checks.

Separated non-filter and filter routing into their own classes.  Adding new non-filter routes only now needs modifications to the new RouteApi class.

Summary of the renaming:
- AuthorizeApi became FilterApi. It is exclusively devoted to filters now.
- AuthRoutingEntry becomes Route to simplify naming.

Changes to AuthTokenTest:
- Previously unsupported HTTP methods were 400, now they are 405. 
- We no longer check text in error responses that flow through the non-filter routes since OpenApi now generates those responses rather than our code.
- Renamed variables to better reflect the new token types instead of "basicToken" etc. Makes the code better but caused a large diff.
- Removed a duplicated variable (basicToken2).

Removed routes and tests for token encryption.

Tested on VM.

Added javadoc throughout.

Note: the `/token` route is the only route that is used in production. Refactored that and commented it to try to make it more clear that it handles POST data in two different shapes (for dummy signing requests and for access token signing requests). They only prop they share is `sub`.


src/main/java/org/folio/auth/authtokenmodule/apis/Api.java
```diff
@@ -0,0 +1,47 @@
...
+  protected static Logger logger;
+
+  protected static void endText(RoutingContext ctx, int code, String msg) {
+    logger.error(msg);
+    ctx.response().setStatusCode(code);
+    ctx.response().putHeader(MainVerticle.CONTENT_TYPE, "text/plain");
+    ctx.response().end(msg);
+  }
+
+  protected static void endJson(RoutingContext ctx, int code, String msg) {
```
julianladisch, line 21: Can you rename msg to json?

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -131,14 +127,13 @@ public static void setUpClass(TestContext context) throws NoSuchAlgorithmExcepti
           if (mainRes.failed()) {
             context.fail(mainRes.cause());
           } else {
+            var tenantAttributes = new JsonObject().put("module_to", "mod-authtoken-1.0.0");
+            initializeTenantForTokenStore(tenant, tenantAttributes);
             async.complete();
```
julianladisch, line 132: This should be futurized to avoid these nested if-clauses.

```
vertx.deployVerticle(...)
.compose(x -> vertx.deployVerticle(...))
.onComplete(context.asyncAssertSuccess(x -> {
  var tenantAttributes = new JsonObject().put("module_to", "mod-authtoken-1.0.0");
  initializeTenantForTokenStore(tenant, tenantAttributes);
}));
```

The logger.info is not needed.

src/main/resources/headers/okapi-token.yaml
```diff
@@ -0,0 +1,7 @@
+in: header
+name: X-Okapi-Token
+description: Okapi Token
+# The /_/tenant route will not have this header.
+required: false
```
julianladisch, line 5: `required: false` means optional.
If some routes _require_ the token an okapi-token-required.yaml with `required: true` should be used.
If `/_/tenant` doesn't use a Token at all the okapi-token.yaml should be omitted.

src/main/resources/headers/okapi-token.yaml
```diff
@@ -0,0 +1,7 @@
+in: header
+name: X-Okapi-Token
+description: Okapi Token
+# The /_/tenant route will not have this header.
```
julianladisch, line 4: Which .yaml file contains the `/_/tenant` route?

src/main/resources/headers/okapi-token.yaml
steveellis: Good catch. Yes agree.

src/main/resources/headers/okapi-token.yaml
steveellis: Also good catch. No file does. I need to add it.

src/main/java/org/folio/auth/authtokenmodule/apis/Api.java
steveellis: Sure no problem.

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -0,0 +1,110 @@
+openapi: 3.0.0
```
julianladisch, line 1: Is this ready to add a Jenkinsfile entry that lints and published the openapi files?
https://dev.folio.org/guides/jenkinsfile/#back-end-modules

steveellis: Yes I think so. Will modify the Jenkins file.

src/main/resources/headers/okapi-token.yaml
steveellis: I think we may not want to replicate the yaml, but instead perhaps just note where to find the yaml for the /_/tenant route. It's here: https://github.com/folio-org/folio-vertx-lib/blob/master/core/src/main/resources/openapi/tenant-2.0.yaml. Duplicating may lead to bugs (not even sure if it would work). If we can make it clear in the comments what is going on (it's passing through the request to a handler in a dependent library folio-vertx-lib).

src/main/resources/headers/okapi-token.yaml
steveellis: See below for comment on how I propose we handle the dependency for `/_/tenant`.  But one side effect of this is that we'll need to leave this `required: false`. The other module that I'm aware of that uses folio-vertx-lib [also left it as](https://github.com/folio-org/mod-eusage-reports/blob/8cdf125331cd6b6fd31428d4b283062b4b1d7f1b/src/main/resources/openapi/headers/okapi-tenant.yaml#L4) `required: false` probably for this reason.

dcrossleyau:
The headers and schemas directories need to be inside the openapi directory.

steveellis:
> The headers and schemas directories need to be inside the openapi directory.

Thanks David! Was just puzzling over my sad jenkins report and see what you mean. :)

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Done

src/main/java/org/folio/auth/authtokenmodule/apis/Api.java
steveellis: Done

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -0,0 +1,113 @@
...
+info:
+  title: mod-authtoken API
+  version: v1
+paths:
+  # NOTE This module relies on the its folio-vertx-lib dependency to handle the /_/tenant
+  # route. Please see its yaml specification for /_/tenant here:
+  # https://github.com/folio-org/folio-vertx-lib/blob/master/core/src/main/resources/openapi/tenant-2.0.yaml
+  /token:
```
steveellis, line 9: @julianladisch  see what you think. I'm thinking this comment is probably enough to address the handing of `/_/tenant`.

src/main/resources/openapi/token-1.0.yaml
steveellis: Added. After moving the directories around the linting seems to be working. See what you think.

dcrossleyau: Yes. I viewed the Artifacts apiLintReport. All good.

src/main/resources/openapi/token-1.0.yaml
julianladisch: I agree.
The official documentation link is https://dev.folio.org/reference/api/#folio-vertx-lib

src/main/resources/openapi/headers/okapi-tenant.yaml
```diff
@@ -0,0 +1,6 @@
+in: header
```
julianladisch, line 1: This overwrites the file from https://github.com/folio-org/folio-vertx-lib/tree/master/core/src/main/resources/openapi/headers
If the file is the same there is no need to overwrite it. If it is different a different file name should be used, for example okapi-tenant-required.yaml.
The same for okapi-token.yaml and okapi-url.yaml.

src/main/resources/openapi/token-1.0.yaml
steveellis: Ok great, thanks Julian. And thanks for pointing out where the generated docs end up. I'll use that ink instead.

Jenkinsfile
```diff
@@ -1,15 +1,21 @@
...
+  apiDirectories = 'src/main/resources/openapi'
+  apiExcludes = 'headers parameters'
+
   doDocker = {
     buildJavaDocker {
-      publishMaster = 'yes'
-      healthChk = 'no'
+      publishMaster = true
+      healthChk = true
+      healthChkCmd = 'curl -sS --fail -o /dev/null  http://localhost:8081/admin/health || exit 1'
```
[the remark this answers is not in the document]
steveellis: Ok thanks. Changing to wget.

src/main/java/org/folio/auth/authtokenmodule/apis/Api.java
```diff
@@ -0,0 +1,47 @@
...
+
+  protected void handleTokenValidationFailure(Throwable h, RoutingContext ctx,
+      String msg, String unexpectedExceptionMsg) {
+    if (h instanceof TokenValidationException) {
+      var e = (TokenValidationException)h;
+      logger.error("{}: {}", msg, e.toString());
+      endText(ctx, e.getHttpResponseCode(), msg);
+      return;
+    }
+    logger.error("{}: {}", unexpectedExceptionMsg, h.toString());
```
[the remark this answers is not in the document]
steveellis: I overrode `toString` in `TokenValidationException` [here](https://github.com/folio-org/mod-authtoken/blob/9ff15dc3f67343f742340527e51f4b2d06660156/src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationException.java#L30) in order to provide the underlying cause to logging, but can see that if I call logger per your suggestion the cause is provided in the stack trace. So I removed the override. See what you think.

src/main/resources/openapi/headers/okapi-tenant.yaml
steveellis: Ok thanks. I removed the yaml file for token (since isn't required and the underlying folio-vertx-yaml handles that case). I renamed the two files (url and tenant) to be `-required` since they change that underlying behavior.

steveellis:
@julianladisch I mentioned this on slack but thought I would put more info here.

I'm thinking about rolling back [the last commit](https://github.com/folio-org/mod-authtoken/pull/122/commits/07599bc868ed1cffcc18f238941c14050660c2b6). In a [prior commit](https://github.com/folio-org/mod-authtoken/pull/122/commits/38af4c4305101c04183ce436c26de4b867e6646c) I removed the encryption endpoints. I think that is fine since they were in use and saved me having to implement them in OpenApi. We discussed that in our standup last week, but with this latest commit I removed encryption from the RT lifecycle.

I think removing this particular use of encryption might be a bit premature. We may end up wanting to encrypt RTs in storage for example, and if we do that we could also exchange them in an encrypted form as well. In any case we will be implementing the full RT lifecycle in a subsequent ticket it might be best to wait on this change until then.

From the [OAuth2 spec doc](https://datatracker.ietf.org/doc/html/rfc6749#section-10.4):

> Refresh tokens MUST be kept confidential in transit and storage.

There is more than one way of keeping tokens confidential in storage but encrypting the tokens would be one.

adamdickmeiss (approved):
Looks good and appears to work fine at least with my local setup.

PR #123, opened 2022-03-30
MODAT-127: break up unit tests

steveellis:
As discussed the `AuthTokenTest` class had one test method which contained many dozens of rest assured tests. This makes figuring out which part of the test is breaking harder than it should be. This PR attempts to remedy that by dividing that one big test method up into several smaller methods.

Where it was possible to make the test have only one request and response I did that. Some sequences of requests and responses were dependent on each other. Where that was the case I combined the sequence into a single method. When tests have more than one request/response I left in the `log.info` statements. When a test contained a single request/response I removed the `log.info` statement and named the method something equivalent to the logging statement.

So for example `logger.info("POST empty token signing request")` becomes `public void testEmptyTokenSigningRequest()`.

### DIFF
--- a/src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
+++ b/src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
@@ -56,6 +56,7 @@ public class AuthTokenTest {
   private static String userUUID = "007d31d2-1441-4291-9bb8-d6e2c20e399a";
   private static String accessToken;
   private static String moduleToken;
+  private static String dummyToken;
   private static String badAccessToken;
   private static String accessToken404;
   private static String inactiveToken;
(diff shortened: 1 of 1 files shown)


adamdickmeiss (approved):
LGTM

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -654,286 +517,448 @@ public void testHttpEndpoints() throws JOSEException, ParseException {
...
+              PermsMock.SYS_PERM_SUB_01 + "," + PermsMock.SYS_PERM_SUB_02)
+          .get("/testsysperm")
+          .then()
+          .statusCode(202)
+          .extract().response();
+
+      String headers = r.getHeader("X-Okapi-Permissions");
+      assertTrue(headers.contains(PermsMock.SYS_PERM_SUB_01));
+      assertTrue(headers.contains(PermsMock.SYS_PERM_SUB_02));
+    }
```
julianladisch, line 961: Code coverage of /src/main/java decreases from 88,9 % to 88,2 %.
2022-03-31-AuthTokenTest-a8311fe
2022-03-31-AuthTokenTest-ea9579d

steveellis: Thanks Julian for thinking to run the coverage report. I had indeed missed one test case, which is now called `testSigningRequestGoodAccessTokenGoodPayload`. After adding this case the coverage report from Idea looks a bit better (using Idea and the JaCoCo coverage runner):

On master:
image

After adding missing case:
image

Here's what it looked like before the change:

image

I have gone through now a second time and checked that each case in both versions is there. How does it look in Eclipse?

julianladisch: Well done, code coverage of /src/main/java is now exactly the same as before in Eclipse.

julianladisch (approved):
Looks good, thanks!

PR #124, opened 2022-03-31
MODAT-128: jwt.signing.key hint on BadSignatureException

julianladisch:
As a sysop I would like to read some hint in the log how to fix the configuration when "Invalid token" might be caused by a wrong jwt.signing.key.

Use case 1:

Given that jwt.signing.key is not set and mod-authtoken is restarted

When using a token that has been issued before the restart that causes an "Invalid token" error

Then the log should indicate this it might be caused by the mod-authtoken restart.

Use case 2:

Given that a cluster of multiple instances of mod-authtoken has been configured with different jwt.signing.key values.

When using a token that has been issued by an instance with different key and getting an "Invalid token" error.

Then the log should indicate this it might be caused by different keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ The expiration time is hard-coded:
 
 mod-authtoken supports a number of command line options as system properties, set by passing `-D<property.name>=<value>` to the jar when loading.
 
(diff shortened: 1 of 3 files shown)


steveellis (approved):
This is a nice clarification.

PR #125, opened 2022-04-27
MODAT-112: implement access token expiration

steveellis:
- Created new endpoints to handle new refresh token behavior (`/token/sign` and `token/refresh`).
- Added new code to return both AT and RT in each response.
- Wired up endpoints to backend RT checks in the RefreshTokenStore.
- Wrote tests for new methods. Unit tests that run against the old endpoints have been given a `_legacy` suffix on the method name to make removing them easier when the time comes.
- Added `exp` claim to AT. It's now checked in `validate`.
- Created new token type `LegacyAccessToken`. This is what is returned from the legacy `/token` endpoint.
- Left in legacy `/refreshtoken` endpoint so as to not break mod-login. All code should be compatible with existing modules since all the old endpoints are still supported.
- Bumped version from 2.0 to 3.0. Added new endpoints to mod descriptor.
- Made the RefreshTokenStore token store return TokenValidationException objects. This made the api more consistent and allowed for validation failures that occur as a result of checking the stores to bubble up the http response code we want.
- Did same for ApiTokenStore to be consistent.
- Updated unit tests to check for the exception type (through a cast) and test for the response code.
- For refresh tokens, made the `TokenValidationException` message more clear about the different cases that could cause the exception to occur.
- Moved some methods around (notably token parse) so that they appear in more natural places in relation to other code.


src/main/java/org/folio/auth/authtokenmodule/apis/Route.java
```diff
@@ -91,7 +91,7 @@ public boolean handleRoute(RoutingContext ctx, String authToken, String moduleTo
       }
     }
     if (!allFound) {
-      logger.error("Insufficient permissions to access endpoint {}", endpoint);
+      logger.error("Insufficient permissions to access endpoint {} {}", endpoint, requiredPermissions);
```
julianladisch, line 94:
```suggestion
      logger.error("Insufficient permissions to access endpoint {}: {}", endpoint, requiredPermissions);
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -18,8 +18,13 @@
...
           "methods": [ "POST" ],
-          "pathPattern": "/refresh",
-          "permissionsRequired": []
+          "pathPattern": "/token/sign",
+          "permissionsRequired": [ "auth.signtoken" ]
+        },
+        {
+          "methods": [ "POST" ],
+          "pathPattern": "/token/refresh",
+          "permissionsRequired": [ "auth.signrefreshtoken" ]
```
julianladisch, line 27: This solution alters

      "id": "authtoken",
      "version": "2.0",

to become authtoken version 2.1.

A different approach simply adds a new

      "id": "authtoken2",
      "version": "1.0",

(or any other id name like "token") and leaves "authtoken" "2.0" as is.

descriptors/ModuleDescriptor-template.json
```diff
@@ -18,8 +18,13 @@
         },
         {
           "methods": [ "POST" ],
-          "pathPattern": "/refresh",
```
julianladisch, line 21: Is it intended that this API is removed at this time?

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -71,13 +82,16 @@ public Future<Router> createRouter(Vertx vertx) {
     return RouterBuilder.create(vertx, "openapi/token-1.0.yaml")
         .map(routerBuilder -> {
           routerBuilder
-              .operation("refresh")
-              .handler(this::handleRefresh);
+              .operation("token-legacy")
+              .handler(this::handleSignLegacyToken);
+              routerBuilder
```
julianladisch, line 87: indentation

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -150,21 +163,30 @@ private void handleSignToken(RoutingContext ctx) {
       if (isDummyTokenSigningRequest(payload)) {
         logger.debug("Signing request is for a dummy token");
 
-        token = new DummyToken(tenant, payload.getJsonArray("extra_permissions"), username);
+        var dt = new DummyToken(tenant, payload.getJsonArray("extra_permissions"), username);
+        responseObject.put("token", dt.encodeAsJWT(tokenCreator));
+        endJson(ctx, 201, responseObject.encode());
       } else {
```
julianladisch, line 169: Inserting `return;` avoids the else clause and its indentation and is more easy to understand for me:

```
        endJson(ctx, 201, responseObject.encode());
        return;
      }
      logger.debug("Signing request is for an access token");

      String userId = payload.getString(USER_ID);
```

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -150,21 +163,30 @@ private void handleSignToken(RoutingContext ctx) {
...
 
-        String userId = payload.getString("user_id");
-        token = new AccessToken(tenant, username, userId);
+        String userId = payload.getString(USER_ID);
+
+        // Generate the access token.
+        var at = new AccessToken(tenant, username, userId);
+        responseObject.put(ACCESS_TOKEN, at.encodeAsJWT(tokenCreator));
+
+        // Generate the refresh token.
```
julianladisch, line 178: The comments

    // Generate the access token.

and

    // Generate the refresh token.

are not needed because the code sections are short and every Java programmer can easily get this information from the code (`new AccessToken`, `new RefreshToken`).

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1070,11 +1382,23 @@ public void testStoreRefreshTokenSingleUse(TestContext context) {
...
+          assertThat(((TokenValidationException)b).getHttpResponseCode(), is(401));
+          assertThat(b.getMessage(), containsString("leaked"));
+          assertThat(b.getMessage(), containsString("revoked"));
+        }))
         .compose(a -> ts.checkTokenNotRevoked(rt3))
-        .onComplete(context.asyncAssertFailure(b -> assertThat(b.getMessage(), containsString("revoked"))));
+        .onComplete(context.asyncAssertFailure(b -> {
+          assertThat(((TokenValidationException)b).getHttpResponseCode(), is(401));
+          assertThat(b.getMessage(), containsString("leaked"));
+          assertThat(b.getMessage(), containsString("revoked"));
```
julianladisch, line 1400: Can you extract this duplicated code into a method like `assertLeakedOrRevoked(Throwable t)`?

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -150,21 +163,30 @@ private void handleSignToken(RoutingContext ctx) {
...
         permissionsSource.clearCacheUser(userId, tenant);
-      }
-
-      logger.debug("Successfully created and signed token");
 
-      JsonObject responseObject = new JsonObject().put("token", token.encodeAsJWT(tokenCreator));
-      endJson(ctx, 201, responseObject.encode());
+        // Save the RT to track one-time use.
+        new RefreshTokenStore(vertx, tenant).saveToken(rt).onComplete(x ->
+          endJson(ctx, 201, responseObject.encode()));
```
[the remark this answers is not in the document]
julianladisch: Can you add a unit tests for the failure case?

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -206,46 +256,9 @@ protected void validateCommon(HttpServerRequest request) throws TokenValidationE
...
-      case ApiToken.type:
-        return new ApiToken(sourceToken, claims);
-      case DummyToken.type:
-        return new DummyToken(sourceToken, claims);
-      case ModuleToken.type:
-        return new ModuleToken(sourceToken, claims);
-      default:
-        throw new TokenValidationException("Unable to parse token", 400);
-    }
+  protected boolean tokenHasExpired(JsonObject tokenClaims) {
```
julianladisch, line 259: `tokenHasExpired` -> `isExpired`

src/main/java/org/folio/auth/authtokenmodule/tokens/RefreshToken.java
```diff
@@ -84,17 +85,12 @@ protected Future<Token> validateContext(TokenValidationContext context) {
...
       var e = new TokenValidationException("Attempt to refresh with expired refresh token", 401);
       return Future.failedFuture(e);
     }
 
-    // TODO Check storage to ensure that token has not yet been used.
-    // TODO If the token has been used, revoke all RTs for this user_id.
-
-    return Future.succeededFuture(this);
+    var refreshTokenStore = (RefreshTokenStore)context.getTokenStore();
+    return refreshTokenStore.checkTokenNotRevoked(this).compose(x -> Future.succeededFuture(this));
```
julianladisch, line 94:
```suggestion
    return refreshTokenStore.checkTokenNotRevoked(this).map(this);
```

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -206,46 +256,9 @@ protected void validateCommon(HttpServerRequest request) throws TokenValidationE
...
+    Long nowTime = Instant.now().getEpochSecond();
+    Long expiration = tokenClaims.getLong("exp");
+    return nowTime > expiration;
```
julianladisch, line 262:
```suggestion
    return nowTime >= expiration;
```

descriptors/ModuleDescriptor-template.json
steveellis: I think yes this API can be removed because it was never used in code and removing it means one less thing for us to maintain. On the other hand, the `/refreshtoken` route (to get a signed refresh token but still - "legacy old code") is currently called by mod-login, although the token it gets is not used, it is still called, so I left it in so as to not break mod-login.

Originally I had also removed `/refreshtoken` as well but @adamdickmeiss  advised keeping it in so as to not be a breaking change and I agreed since I'd have to make changes to mod-login before merging this if I took it out. So even though this removes the `/refresh` route it can still be merged.

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
steveellis: Agree. Eliminating else is a good thing.

descriptors/ModuleDescriptor-template.json
julianladisch: In theory it's a breaking change, in practice nobody notices.

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -179,43 +201,95 @@ private void handleRefresh(RoutingContext ctx) {
...
-          JsonObject responseObject = new JsonObject().put("token", at);
-          endJson(ctx, 201, responseObject.encode());
+          JsonObject responseObject = new JsonObject().put(ACCESS_TOKEN, at);
+
+          String address = ctx.request().remoteAddress().host();
+          var rt = new RefreshToken(tenant, username, userId, address);
+          responseObject.put(REFRESH_TOKEN, rt.encodeAsJWE(tokenCreator));
+
+        // Save the RT to track one-time use.
+        new RefreshTokenStore(vertx, tenant).saveToken(rt).onComplete(x ->
```
julianladisch, line 228: Is it possible to use the usual pattern with .compose and error handling at the end avoiding the try-catch?

```
Token.validate(context)
.compose(token -> {
  ...
  return new RefreshTokenStore(vertx, tenant).saveToken(rt)
      .map(responseObject);
})
.onSuccess(responseObject -> endJson(ctx, 201, responseObject.encode()));
.onFailure(e -> ...);
```

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
julianladisch, line 259: If it is not static it should check the `protected JsonObject claims` of this Token.
If it takes a toeknClaims parameter it should be static.

src/main/resources/openapi/schemas/tokenResponse.json
```diff
@@ -0,0 +1,19 @@
+{
+  "description": "The signed JWT token based on the payload",
```
julianladisch, line 2: description doesn't match properties

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -559,11 +584,12 @@ public void testSigningRequestGoodDummyTokenGoodPayload() {
           .post("/token")
           .then()
           .statusCode(201).contentType("application/json").extract().path("token");
-      assertThat(new OkapiToken(token).getUsernameWithoutValidation(), is(payloadDummySigningReq.getString("sub")));
-    }
+      var td = (DummyToken)Token.parse(token, tokenCreator);
+      assertThat(td.getClaims().getString("sub"), is(payloadDummySigningReq.getString("sub")));
```
julianladisch, line 588: A new Token method `public String getClaim(String)` could simplify it.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -575,11 +601,13 @@ public void testSigningRequestGoodAccessTokenGoodPayload() {
           .post("/token")
           .then()
           .statusCode(201).contentType("application/json").extract().path("token");
-      assertThat(new OkapiToken(token).getUsernameWithoutValidation(), is(payloadSigningRequest.getString("sub")));
+       var lat = (LegacyAccessToken)Token.parse(token, tokenCreator);
+       assertThat(lat.getClaims().getString("sub"), is(payloadSigningRequest.getString("sub")));
+       assertNull(lat.getClaims().getString("exp"));
```
julianladisch, line 606: indentation

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -634,61 +662,237 @@ public void testSigningRequestGoodTokenBadPayload3() {
...
+            .header("Content-type", "application/json")
+            .header("X-Okapi-Permissions", "[\"" + getMagicPermission("/token/sign") + "\"]")
+            .body(new JsonObject().put("payload", new JsonObject().put("x", 1)).encode())
+            .post("/token/sign")
+            .then()
+            .statusCode(400);
+      }
+
+      @Test
+      public void testRefreshTokenBadPayload3() {
```
julianladisch, line 884: 3?

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
steveellis: Done. See new `returnToken` method.

src/main/java/org/folio/auth/authtokenmodule/tokens/RefreshToken.java
steveellis: Done. See new `returnToken` method.

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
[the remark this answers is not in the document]
steveellis: Done, although the unit tests for the failure case of saveToken are implicit.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Agreed. It's an instance method and has access to the claims. Removing param.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Fixed

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -801,6 +1006,107 @@ public void testRefreshToken() throws JOSEException, ParseException {
...
+          .then()
+          .statusCode(401).body(is("Invalid token"));
+    }
+
+    @Test
+    public void testAllTokensRevokedAfterOneTokenIsRevoked() throws JOSEException, ParseException {
+      logger.info("Create three refresh tokens to simulate multiple logins");
+      var tokens = new ArrayList<String>();
+      for (int i = 0; i < 3; i++) {
+      String refreshToken = given()
```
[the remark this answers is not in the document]
steveellis: Fixed

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Not sure how that 3 crept in there. No need for it. Fixed

src/main/resources/openapi/schemas/tokenResponse.json
steveellis: Agreed. Changed to better describe how the schema is used. Its either a `token` (when it's a dummy signing request) or both a `refreshToken` and an `accessToken` when it is a "regular" signing request.

descriptors/ModuleDescriptor-template.json
steveellis: Ok added second group. Decided `authtoken2` was a good enough name. Also think it is nice to represent on the top level that there's an "old way" and a "new way" and that there is no overlap.

julianladisch (approved):
Looks good, thanks!

descriptors/ModuleDescriptor-template.json
```diff
@@ -15,11 +15,22 @@
           "methods": [ "POST" ],
           "pathPattern": "/refreshtoken",
           "permissionsRequired": [ "auth.signrefreshtoken" ]
+        }
+      ]
+    },
+    {
+      "id": "authtoken2",
```
adamdickmeiss, line 22: Given that permission `auth.signtoken` is now shared between two interfaces I would have preferred to simply bump authtoken to 2.1.

adamdickmeiss (approved):

descriptors/ModuleDescriptor-template.json
steveellis: We went with the 2 groups so that we have some way to represent on the top level how there are now two different interfaces.

steveellis:
Tested on VM before merging.

PR #129, opened 2022-06-28
MODAT-133: Finalize refresh token backend

steveellis:
See the ticket for details. This should have the new logout endpoints implemented, but no tests yet for them. Also added token exp to the response. Tests for those are there.


descriptors/ModuleDescriptor-template.json
```diff
@@ -31,6 +31,16 @@
...
           "permissionsRequired": [ "auth.signrefreshtoken" ]
+        },
+        {
+          "methods": [ "POST" ],
+          "pathPattern": "/token/logout",
+          "permissionsRequired" : [ ]
+        },
+        {
+          "methods": [ "DELETE" ],
+          "pathPattern": "/token/logout-all",
```
julianladisch, line 42: For me this looks odd.

If there's POST /token/logout I would expect POST /token/logout-all.

https://datatracker.ietf.org/doc/html/rfc2616#section-9.7 "The DELETE method requests that the origin server delete the resource identified by the Request-URI."

Is the resource "logout-all" to be deleted?

steveellis: I agree that the POST to /token/logout is inconsistent with the DELETE to /token/logout-all. My goal was to use DELETE for both operations here like I did mod-login, but ended up not being able to make this happen due the the need for a body to /token/logout.

Let me change these to both be POST here. Let me know in the mod-login PR if  you'd like to have those be POST as well.

julianladisch: Yes, use POST there as well.

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -228,9 +238,9 @@ private boolean isDummyTokenSigningRequest(JsonObject payload) {
 
   private void handleRefresh(RoutingContext ctx) {
     try {
-      String content = ctx.getBodyAsString();
+      String content = ctx.body().asString();
```
julianladisch, line 241: ctx.body().asJsonObject()

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -253,14 +263,51 @@ private void handleRefresh(RoutingContext ctx) {
     }
   }
 
+  private void handleTokenLogout(RoutingContext ctx) {
+    try {
+      logger.debug("Called handleTokenLogout");
+      String content = ctx.body().asString();
```
julianladisch, line 269: .asJsonObject()

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -253,14 +263,51 @@ private void handleRefresh(RoutingContext ctx) {
...
+
+      RefreshToken rt = (RefreshToken)Token.parse(encryptedJWE, tokenCreator);
+
+      var refreshTokenStore = new RefreshTokenStore(vertx, tenant);
+      refreshTokenStore.revokeToken(rt)
+          .onSuccess(x -> endNoContent(ctx, 204))
+          .onFailure(e -> handleTokenValidationFailure(e, ctx));
+    } catch (Exception e) {
+      endText(ctx, 500, String.format("Unanticipated exception when handling token logout: %s", e.getMessage()));
+    }
```
julianladisch, line 282: This and the following catch should log the exception and its stacktrace.

I would prefer concise message, format is not needed:

    endText(ctx, 500, "Cannot handle token logout: " + e.getMessage());

src/main/java/org/folio/auth/authtokenmodule/tokens/RefreshToken.java
```diff
@@ -18,6 +18,8 @@ public class RefreshToken extends Token {
    */
   public static final String TYPE = "refresh";
 
+  private static final long EXPIRATION_SECONDS = 604800;
```
julianladisch, line 21: Can you add a comment like /** 604800 seconds = 7 days */

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -107,6 +106,40 @@ paths:
...
+        "400":
+          $ref: "#/components/responses/trait_400"
+        "500":
+          $ref: "#/components/responses/trait_500"
+  /token/logout-all:
+    parameters:
+      - $ref: headers/okapi-tenant-required.yaml
+      - $ref: headers/okapi-url-required.yaml
+    post:
+      description: Logout a single token
```
julianladisch, line 134: all tokens

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -107,6 +106,40 @@ paths:
...
+      operationId: token-logout
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/refreshToken"
```
julianladisch, line 120: Who will copy the refresh token from the cookie to the body?
The front-end cannot do this because the front-end cannot read the refresh cookie.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1093,18 +1057,103 @@ public void testAllTokensRevokedAfterOneTokenIsRevoked() throws JOSEException, P
...
-      }
+          .statusCode(401);
+    }
+
+    @Test
+    public void testLogoutAll() throws JOSEException, ParseException {
+      logger.info("Create some refresh tokens for logout all test");
+      var tokens = createListOfRefreshTokens();
+
+      logger.info("POST /token/logout-all (bad method)");
```
julianladisch, line 1133: PUT

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
```diff
@@ -1093,18 +1057,103 @@ public void testAllTokensRevokedAfterOneTokenIsRevoked() throws JOSEException, P
...
+          .header("X-Okapi-Tenant", tenant)
+          .header("X-Okapi-Token", accessToken)
+          .header("X-Okapi-Url", "http://localhost:" + freePort)
+          .header("Content-type", "application/json")
+          .header("X-Okapi-Permissions", "[\"" + getMagicPermission("/token/logout-all") + "\"]")
+          .put("/token/logout-all")
+          .then()
+          .statusCode(405);
+
+      logger.info("DELETE /token/logout-all");
```
julianladisch, line 1144: POST

descriptors/ModuleDescriptor-template.json
steveellis: Ok will change mod-login to POST for the new logout endpoints there as well.

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
steveellis: Done

src/main/resources/log4j2.properties
```diff
@@ -14,6 +14,6 @@ appender.console.name = STDOUT
 appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = %d{HH:mm:ss} [$${FolioLoggingContext:requestid}] [$${FolioLoggingContext:tenantid}] [$${FolioLoggingContext:userid}] [$${FolioLoggingContext:moduleid}] %-5p %-20.20C{1} %m%n
 
-rootLogger.level = info
-rootLogger.appenderRefs = info
+rootLogger.level = debug
+rootLogger.appenderRefs = debug
```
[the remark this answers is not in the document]
steveellis: Fixed

src/main/java/org/folio/auth/authtokenmodule/tokens/RefreshToken.java
steveellis: Yep

src/main/resources/openapi/token-1.0.yaml
steveellis: Mod login is calling this, not the front end. Mod-login has code for parsing the cookie and making the request to mod-login.

src/main/resources/openapi/token-1.0.yaml
steveellis: Right. Fixed.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Fixed.

src/test/java/org/folio/auth/authtokenmodule/AuthTokenTest.java
steveellis: Fixed

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
steveellis: Sure, changed everywhere.

src/main/java/org/folio/auth/authtokenmodule/tokens/RefreshToken.java
steveellis: @julianladisch  I updated readme to refer to the new OpenApi yaml since before the only documentation  of the endpoints were in the Readme. Now we have the yaml. Clarified the yaml  to include the token expirations. Also noticed that the tokenResponse.json wasn't being referred to properly in the  yaml. Fixed that.
    
Also noticed that the tokenResponse.json didn't have the new expiration properties  in it. Fixed that as well.

Finally moved the `tokenResponse` method in `RouteApi` to be below where it is called in both places to make the code easier to follow.

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -107,6 +106,40 @@ paths:
           $ref: "#/components/responses/trait_400"
         "500":
           $ref: "#/components/responses/trait_500"
+  /token/logout:
```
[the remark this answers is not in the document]
steveellis: The logout routes here in mod-authtoken are called by the logout routes in mod-logout. Each has its own purpose. In the case of mod-login, the purpose of the logout routes are to respond with `Set-Cookie` with a time in the past to remove the cookie on the client. The purpose of the logout routes in mod-authtoken are to perform the actual removal of the refresh token from storage to ensure that it cannot be used to request new access tokens.

In other words: mod-login has responsibility for cookies and mod-authtoken has responsibility for storage.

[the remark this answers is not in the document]
marcjohnson-kint: > The logout routes here in mod-authtoken are called by the logout routes in mod-logout. Each has its own purpose. 

I agree that they have different purposes.

To me, they operate at different levels.

> In the case of mod-login, the purpose of the logout routes are to respond with `Set-Cookie` with a time in the past to remove the cookie on the client. The purpose of the logout routes in mod-authtoken are to perform the actual removal of the refresh token from storage to ensure that it cannot be used to request new access tokens. In other words: mod-login has responsibility for cookies and mod-authtoken has responsibility for storage.

Ok.

Could this be named to refer to the tokens being removed or invalidated, rather than logging in / out?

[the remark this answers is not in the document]
steveellis: I'm not against this if you think it would be an improvement. What would you prefer? Could do `/token/invalidate` and 'token/invalidate-all`. That might make it a bit closer semantically to what is happening. I think I would prefer "invalidate" over "remove" since the data deletion seems somewhat lower down whereas "validation/invalidation" is the broader concept and it is a term that is used throughout the source code for this feature. "Revoke" might be another candidate but it doesn't have convenient a opposite like "validate/invalidate".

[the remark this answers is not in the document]
marcjohnson-kint: > I think I would prefer "invalidate" over "remove" since the data deletion seems somewhat lower down whereas "validation/invalidation" is the broader concept and it is a term that is used throughout the source code for this feature. 

I'm good with this reasoning. You are close to the domain and understand the semantic intent of the operation.

I think the term being used throughout the source is a good indicator.

adamdickmeiss (approved):
Appears to work fine and not break compatibility from the tests I've done.

descriptors/ModuleDescriptor-template.json
```diff
@@ -31,6 +31,14 @@
           "methods": [ "POST" ],
           "pathPattern": "/token/refresh",
           "permissionsRequired": [ "auth.signrefreshtoken" ]
+        },
+        {
+          "methods": [ "POST" ],
+          "pathPattern": "/token/logout"
```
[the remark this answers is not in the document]
steveellis: Added

descriptors/ModuleDescriptor-template.json
```diff
@@ -31,6 +31,14 @@
...
+        },
+        {
+          "methods": [ "DELETE" ],
+          "pathPattern": "/token/logout-all"
```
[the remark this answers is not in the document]
steveellis: Same

src/main/resources/openapi/token-1.0.yaml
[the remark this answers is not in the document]
steveellis: Renamed /logout routes to /invalidate.

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -63,7 +65,10 @@ paths:
       - $ref: headers/okapi-tenant-required.yaml
       - $ref: headers/okapi-url-required.yaml
     post:
-      description: Sign token
+      description: |
+        Returns a signed, expiring access token and refresh token. Also returns the expiration
+        of each token in the body of the response.The access token time to live is 10 minutes and
```
julianladisch, line 70: Space after dot.
Reword: The time to live is 10 minutes for the access token and one week for the refresh token.

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -82,13 +87,15 @@ paths:
...
           $ref: "#/components/responses/trait_500"
-
   /token/refresh:
     parameters:
       - $ref: headers/okapi-tenant-required.yaml
       - $ref: headers/okapi-url-required.yaml
     post:
-      description: Get a new refresh token and a new access token
+      description: |
+        May be called to obtain a new refresh token and a new access token. Returns a signed, expiring
```
julianladisch, line 96: `May be called to obtain` is very verbose and could be replaced by `Returns`.
`Returns a signed, expiring access token and refresh token.` This can be removed. Expiration is explained in the next sentence, and the client doesn't need information whether the token is signed or encrypted.

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -82,13 +87,15 @@ paths:
...
+        access token and refresh token. Also returns the expiration of each token in the body of the
+        response. The access token time to live is 10 minutes and the refresh token is one week.
```
julianladisch, line 98: Reword: The time to live is 10 minutes for the access token and one week for the refresh token.

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -102,7 +109,41 @@ paths:
...
+        "400":
+          $ref: "#/components/responses/trait_400"
+        "500":
+          $ref: "#/components/responses/trait_500"
+  /token/invalidate:
+    parameters:
+      - $ref: headers/okapi-tenant-required.yaml
+      - $ref: headers/okapi-url-required.yaml
+    post:
+      description: Logout a single token
```
julianladisch, line 122: Invalidate, not logout

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -102,7 +109,41 @@ paths:
...
+      operationId: token-logout
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/refreshToken"
+        required: true
+      responses:
+        "204":
+          description: Logged out token successfully
```
julianladisch, line 132: Invalidated token successfully

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -102,7 +109,41 @@ paths:
...
+        "400":
+          $ref: "#/components/responses/trait_400"
+        "500":
+          $ref: "#/components/responses/trait_500"
+  /token/invalidate-all:
+    parameters:
+      - $ref: headers/okapi-tenant-required.yaml
+      - $ref: headers/okapi-url-required.yaml
+    post:
+      description: Logout all tokens for a user
```
julianladisch, line 142: Invalidate

src/main/resources/openapi/token-1.0.yaml
```diff
@@ -102,7 +109,41 @@ paths:
...
+      operationId: token-logout-all
+      responses:
+        "204":
+          description: Logged out token successfully
```
julianladisch, line 146: Invalidated all tokens successfully

julianladisch (approved):
Looks good, thanks!

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -22,9 +22,13 @@
  * including a method to parse and validate a token in one step.
  */
 public abstract class Token {
+  public static final String REFRESH_TOKEN = "refreshToken";
+  public static final String ACCESS_TOKEN = "accessToken";
```
julianladisch, line 26: wiki.example.com might set a refreshToken and an accessToken for *.example.com that the browser will send to folio.example.com.

To avoid a name clash the cookie name should be more specific, for example folioRefreshToken and folioAccessToken

src/main/resources/openapi/token-1.0.yaml
steveellis: Done

src/main/resources/openapi/token-1.0.yaml
steveellis: Done

src/main/resources/openapi/token-1.0.yaml
steveellis: Done

src/main/resources/openapi/token-1.0.yaml
steveellis: Reworded

julianladisch (changes requested):
See comments:
* https://github.com/folio-org/mod-login/pull/133/files#r922283979
* https://github.com/folio-org/mod-login/pull/133/files#r922284122
* https://github.com/folio-org/mod-login/pull/133/files#r927947190

src/main/java/org/folio/auth/authtokenmodule/apis/FilterApi.java
```diff
@@ -199,205 +183,201 @@ private void handleAuthorize(RoutingContext ctx) {
...
-      if (ctx.request().headers().contains(XOkapiHeaders.MODULE_PERMISSIONS)) {
-        JsonObject modulePermissions = new JsonObject(ctx.request().headers().get(XOkapiHeaders.MODULE_PERMISSIONS));
-        for (String moduleName : modulePermissions.fieldNames()) {
-          JsonArray permissionList = modulePermissions.getJsonArray(moduleName);
-          String moduleToken;
-          try {
-            moduleToken = new ModuleToken(tenant, username, finalUserId, moduleName, permissionList)
+      permService.expandSystemPermissions(extraPermissions, tenant, okapiUrl, permissionsRequestToken,
+        requestId).onSuccess(expandedPermissions -> {
+        // Instead of storing tokens, let's store an array of objects that each
```
julianladisch, line 188: incomplete sentence

src/main/resources/log4j2.properties
julianladisch, line 18: Debug logging should not be the default.

steveellis:
> See comments:
> 
> * https://github.com/folio-org/mod-login/pull/133/files#r922283979
> * https://github.com/folio-org/mod-login/pull/133/files#r922284122
> * https://github.com/folio-org/mod-login/pull/133/files#r927947190

I'm having trouble understanding what you want to do with this this request for changes. These issues are all in mod-login which calls mod-authtoken. How do they affect mod-authtoken? Mod authtoken only returns the tokens and the TTLs. I can see that you may be asking for a different date format (RFC 1123) rather than 8601. But that is just a guess about what you're looking for. Please let me know what you're asking for here.

SerhiiNosko (approved):

Abdulkhakimov (approved):

PR #131, opened 2022-09-27
Handle legacy tokens - dummy and access MODAT-136

adamdickmeiss:


steveellis (approved):

src/test/java/org/folio/auth/authtokenmodule/TokenTest.java
```diff
@@ -71,4 +75,39 @@ public void tokenIsEncryptedTest() throws TokenValidationException, JOSEExceptio
...
     assertThat(Token.isEncrypted(unencryptedToken), is(false));
   }
+
+  @Test
+  public void noTypeDummy() throws TokenValidationException {
+    JsonObject claims = new JsonObject()
+      .put("dummy", true)
+      .put("tenant", "lib");
+    Token parse = Token.parse(".", claims);
+    assertThat(parse instanceof DummyToken, is(true));
```
julianladisch, line 85: For lines 85 and 93 the idiomatic hamcrest way is

    assertThat(parse, is(instanceOf(DummyToken.class)));

https://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/CoreMatchers.html#instanceOf-java.lang.Class-

skoczko (approved):

PR #135, opened 2023-01-04
Use free port in test

adamdickmeiss:
Use free port as done in all other tests. Thus tests will be able to work on systems where 8081 is already in use.


MikeTaylor (approved):
Seems reasonable.

PR #140, opened 2023-05-26
[MODAT-143] - Provide ability to cross tenant requests for consortia members

Abdulkhakimov:
## Purpose
Enabling the ability for queries between tenants who are members of consortia.

## Approach

- Handling the 'Tenant mismatch' exception during token validation.
- Sending a request to the '/user-tenants' endpoint and checking the record count for a specific tenant.
- Added caching for the '/user-tenants' endpoint based on the tenant.
- If the record count after calling '/user-tenants' is greater than zero, permission is granted for further processing of the request, even in case of a mismatch between tenant names in the headers and claims.

## Learning
[MODAT-143](https://issues.folio.org/browse/MODAT-143)


src/main/java/org/folio/auth/authtokenmodule/tokens/AccessToken.java
```diff
@@ -49,7 +49,7 @@ protected Future<Token> validateContext(TokenValidationContext context) {
     try {
       validateCommon(context.getHttpServerRequest());
     } catch (TokenValidationException e) {
-      return Future.failedFuture(e);
+      return handleCrossTenantRequest(e, context);
```
SerhiiNosko, line 52: please refactor to avoid using exception for handling business cases, I propose to refactor validateCommon(), it should return Future object, it will allow to invoke user service if tenants not matches in validateCommon() and simplify validateContext method

Abdulkhakimov: @SerhiiNosko please review again. thank you

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -230,8 +234,12 @@ public static Token parse(String sourceToken, JsonObject claims) throws TokenVal
    * @throws TokenValidationException Throws this when any validation step fails, but may throw
    * other exceptions as well.
    */
-  protected void validateCommon(HttpServerRequest request) throws TokenValidationException {
+  protected Future<Token> validateCommon(TokenValidationContext context) {
+    Promise<Token> promise = Promise.promise();
+
+    var request = context.getHttpServerRequest();
     try {
+      boolean isNeedToComplete = true;
```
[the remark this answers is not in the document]
Abdulkhakimov: done please check

src/main/java/org/folio/auth/authtokenmodule/UserService.java
```diff
@@ -96,6 +100,47 @@ private Future<Boolean> isActiveUserNoCache(String userId, String tenant, String
...
+    if (value == null) {
+      return isUserTenantNotEmptyNoCache(tenant, okapiUrl, requestToken, requestId);
+    }
+    return Future.succeededFuture(value);
+  }
+
+  private Future<Boolean> isUserTenantNotEmptyNoCache(String tenant, String okapiUrl,
+                                              String requestToken, String requestId) {
+
+    HttpRequest<Buffer> req = client.getAbs(okapiUrl + "/user-tenants");
```
NikitaSedyx, line 115: can we have limit 0 here?

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -251,30 +257,56 @@ protected void validateCommon(HttpServerRequest request) throws TokenValidationE
...
+    return promise.future();
   }
 
   protected boolean tokenIsExpired() {
     Long nowTime = Instant.now().getEpochSecond();
     Long expiration = claims.getLong("exp");
     return nowTime > expiration;
   }
+
+  protected Future<Boolean> isCrossTenantRequest(TokenValidationContext context) {
```
okolawole-ebsco, line 296: please add a comment to this method. It is not clear why isUserTenantNotEmpty signal that the request is a cross tenant request.

src/main/java/org/folio/auth/authtokenmodule/apis/FilterApi.java
```diff
@@ -63,20 +63,18 @@ public FilterApi() {}
    * @param routeApi A reference to the RouteApi object, which this API depends on for route
```
steveellis, line 63: Don't forget to add a new `@param` here for your new param

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
```diff
@@ -58,8 +60,9 @@ public class RouteApi extends Api implements RouterCreator, TenantInitHooks {
    *                     shared among
    *                     all Api classes.
```
steveellis, line 61: Don't forget to update the javadoc with your new param

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -251,30 +257,56 @@ protected void validateCommon(HttpServerRequest request) throws TokenValidationE
...
       }
+
+      String headerTenant = request.headers().get(XOkapiHeaders.TENANT);
+      if (!claims.getString("tenant").equals(headerTenant)) {
+        isCrossTenantRequest(context)
+          .compose(aResult -> Boolean.TRUE.equals(aResult) ? Future.succeededFuture(this)
+            : Future.failedFuture(new TokenValidationException(TENANT_MISMATCH_EXCEPTION_MESSAGE, 403)))
+          .onComplete(promise);
+      } else {
+        promise.complete(this);
```
steveellis, line 278: I think we still need the check of `claims.getString("tenant").equals(headerTenant)` don't we when it is not a cross tenant request? Otherwise a token could be used to request another tenant.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Agree, but it would be preferable to add javadocs to all the new methods (following the practice of other method in  this module).

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -230,7 +233,10 @@ public static Token parse(String sourceToken, JsonObject claims) throws TokenVal
    * @throws TokenValidationException Throws this when any validation step fails, but may throw
    * other exceptions as well.
    */
-  protected void validateCommon(HttpServerRequest request) throws TokenValidationException {
+  protected Future<Token> validateCommon(TokenValidationContext context) {
```
steveellis, line 236: Is `isCrossTenantRequest` really in scope for _every_ token type or is it _only_ in scope for `LegacyAccessToken` and `AccessToken`? Although it might appear to work even if you have this check being made in every request, I'm not sure if it is "correct" for the other token types, especially `DummyToken` and the newer token types that aren't released yet.

The idea with `validateCommon` is that only the items that are shared by every token type be placed here. 

The idea with these token type classes is that each type have responsibility performing the validation of its type in the implementation of the `validate` method. If the two access token types (`LegacyAccessToken` and the newer unreleased `AccessToken`) share validation logic perhaps create a base abstract class that these two can inherit from with that shared logic.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
okolawole-ebsco: I believe this code branch is for when there is a tenant id match. check if `claims.getString("tenant").equals(headerTenant)` is already done at the beginning of the if for every request. If there is a tenant mismatch and it is not a cross tenant request, this failed future is returned `Future.failedFuture(new TokenValidationException(TENANT_MISMATCH_EXCEPTION_MESSAGE, 403))` which would yield the same exception as the OLD code that was replaced.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
NikitaSedyx: ModuleToken is in this list as well, DummyToken probably not

SerhiiNosko: I think checking is tenant mismatch - it is core check the same as checking id userId is matched that already contained in validateCommon method [Link](https://github.com/folio-org/mod-authtoken/blob/master/src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java#L265) , and if we remain userId validation in common method and move token validation to subclasses - it will look inconsistent as for me.
Also when adding new token type author can forget about token mismatch validation that can lead to some security hall, because tenant check - it is core validation check

SerhiiNosko: And also before this PR tenant mismatch was checking for all token types, because it was located in validateCommon, and if we are going to move this logic to only particular subclasses - we will not be safe to make this change , because we will broke existing validation contracts for exisitng token types

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
Abdulkhakimov: Done

src/main/java/org/folio/auth/authtokenmodule/apis/FilterApi.java
Abdulkhakimov: done

src/main/java/org/folio/auth/authtokenmodule/apis/RouteApi.java
Abdulkhakimov: done

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -226,11 +229,14 @@ public static Token parse(String sourceToken, JsonObject claims) throws TokenVal
 
   /**
    * Validate all the things that tokens have in common.
-   * @param request The http request context where the token is being provided.
+   * @param context The context for the token validation.
    * @throws TokenValidationException Throws this when any validation step fails, but may throw
```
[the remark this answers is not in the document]
Abdulkhakimov: removed the line

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: Every token type implementation has an abstract `validateContext` method which is called by `Token.validate`. This is where logic that is unique to a given token type belongs. Since `validateContext` is called for every request where a token is in scope, putting validation logic there is "safe". Only if logic is shared by _every_ token type it should go in `validateCommon`. Every `Token` implementation is responsible for its own validation through its `validateContext` implementation.

We shouldn't put this new code in `validateCommon` because it convenient but because it is correct. To me, if this new check of the users table isn't needed by every token type, putting it in `validateCommon` is what could break the contract.

Do I know for a fact that this new logic isn't needed by every token type? I don't, but I suspect that they don't. I think being sure about this should be part of this change. Its easy enough to see the lifecycle of each token type. Ask yourself, will your ECR changes require the user table check for each type type? If you're not completely sure, its safer to change only the token types you are sure of (the access tokens?), and then see what breaks by deploying mod-authtoken in a full system, rather than changing every token type "just in case".

SerhiiNosko: okay, we will rewrite this

Abdulkhakimov: @steveellis please review latest changes. thank you

steveellis: Ok giving token types the ability to opt out of the logic with the override seems fine. It looks like `ModuleToken` and `DummyToken` will need the new logic based on how they are being used.

NikitaSedyx: yes, basically `ModuleToken` and `DummyToken ` should be cross-tenant as well

src/main/java/org/folio/auth/authtokenmodule/UserService.java
```diff
@@ -28,7 +29,10 @@ public UserService(Vertx vertx, int cacheInSeconds, int purgeCacheInSeconds) {
     cachePeriod = cacheInSeconds * 1000L;
 
     // purge cache periodically
-    vertx.setPeriodic(purgeCacheInSeconds * 1000L, id -> cache = new ConcurrentHashMap<>());
+    vertx.setPeriodic(purgeCacheInSeconds * 1000L, id -> {
+      cache = new ConcurrentHashMap<>();
```
steveellis, line 33: Maybe rename this variable `activeUserCache` since it is no longer the only cache.

steveellis (approved):

src/main/java/org/folio/auth/authtokenmodule/UserService.java
SerhiiNosko: Renamed, @steveellis can you please review this PR also: https://github.com/folio-org/mod-users-bl/pull/161 
It is very important for us, because user in institutional tenant could not setup password and so could not login using central tenant UI, we already discussed  this approach, there is no any security violations in exposting tenant to reset password link

steveellis: You're all set @SerhiiNosko. Approved the mod-users-bl one too.

SerhiiNosko: thank you

PR #143, opened 2023-07-17
[MODAT-148]. Allow cross tenant requests only when special system property variable presented

SerhiiNosko:
https://issues.folio.org/browse/MODAT-148

**Overview:**
Currently cross tenant requests allowed only in case when core mod-users /user-tenants response has total > 0 , it means that corresponding tenant added to consortium and we allow cross-tenant request. One more level of protection will be added by using a new system property:
'allow.cross.tenant.requests'. If this propertyis not set - cross tenant requests will be denied, in another case - standard checking of /user-tenant endpoint should be done as it doing now.

**Implementation Approach:**
ValidateTenantMismatch method is adjusted to look into presence of 'allow.cross.tenant.requests' [Code](https://github.com/folio-org/mod-authtoken/blob/master/src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java#L292)

If system property is missed - request will be denied or check of /user-tenants endpoint will be done as before.

### DIFF
--- a/src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationContext.java
+++ b/src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationContext.java
@@ -63,6 +63,7 @@ public class TokenValidationContext {
 
   /**
    * Return flag is cross tenant requests allowed, it uses system property 'allow.cross.tenant.requests'.
+   *
    * @return true if system property 'allow.cross.tenant.requests' equals to 'true' or false if
    * it is not set or equals to 'false'
    */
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/auth/authtokenmodule/tokens/TokenValidationContext.java
```diff
@@ -63,17 +79,15 @@ public TokenValidationContext(HttpServerRequest httpServerRequest,
     this.tokenCreator = tokenCreator;
     this.tokenToValidate = tokenToValidate;
     this.userService = userService;
+    this.allowCrossTenantRequests = BooleanUtils.toBoolean(System.getProperty("allow.cross.tenant.requests"));
```
okolawole-ebsco, line 82: if there are other system properties defined in mod-authtoken and there is a central place where they are all defined, it will be good to include this line in that central place if it exists.

SerhiiNosko: system properties are defined in places where they need to be used, there is no central place, as option to improve I can create class with constants for all system variable names

okolawole-ebsco: if they are defined where needed, you can leave as is.

Dmitriy-Butramyou (approved):

steveellis (approved):

PR #145, opened 2023-08-21
MODAT-149. Make RTR compatible with ECS

SerhiiNosko:
https://issues.folio.org/browse/MODAT-149

The 2 changes made here:

- Remove returning of failed future for refresh token when tenantId in headers do not match with tenantId in token. Instead common validation method used that allows cross tenant requests when all conditions met
- Refactor RefreshToken to use tenantId from token instead of headers to make it compatible with ECS when tenantId in header and token can be different, in this case need to use tenantId from token claims

### DIFF
--- a/src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
+++ b/src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
@@ -82,15 +82,41 @@ public abstract class Token {
   public static Future<Token> validate(TokenValidationContext context) {
     Token token;
     try {
-      token = parse(context.getTokenToValidate(), context.getTokenCreator());
+      token = validateTokenContent(context.getTokenToValidate(), context.getTokenCreator());
     } catch (TokenValidationException e) {
       return Future.failedFuture(e);
-    } catch (Exception e) {
-      return Future.failedFuture(new TokenValidationException("Unexpected token parse exception", e, 500));
     }
 
     // Call the validateContext implementation of the underlying token type (AccessToken,
     // RefreshToken, etc.). See those classes for the validation logic specific to each type.
+    return validateTokenType(token, context);
+  }
+
+  /**
+   * Validates if token payload is correct and parsed successfully.
+   * @param tokenToValidate the token content payload to validate
+   * @param tokenCreator the token creator
+   * @return token if token content parsed successfully
+   * @throws TokenValidationException if token content could not be parsed
+   */
+  public static Token validateTokenContent(String tokenToValidate, TokenCreator tokenCreator) throws TokenValidationException {
+    try {
+      return parse(tokenToValidate, tokenCreator);
+    } catch (TokenValidationException e) {
+      throw e;
+    } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


Abdulkhakimov (approved):

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -82,15 +82,41 @@ public String getTenant() {
...
+   * @param tokenToValidate the token content payload to validate
+   * @param tokenCreator the token creator
+   * @return token if token content parsed successfully
+   * @throws TokenValidationException if token content could not be parsed
+   */
+  public static Token validateTokenContent(String tokenToValidate, TokenCreator tokenCreator) throws TokenValidationException {
+    try {
+      return parse(tokenToValidate, tokenCreator);
+    } catch (TokenValidationException e) {
+      throw e;
```
steveellis, line 106: I don't feel like this method is necessary. I like the previous code better (probably because I wrote it...) because you don't have to rethrow the same exception. The idea was that `parse` would consolidate all exceptions into the one type of `TokenValidationException` and if that type didn't get caught you would at least have a log message that indicated it was something not expected.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
```diff
@@ -82,15 +82,41 @@ public String getTenant() {
...
+  /**
+   * Validates token type, each underlying token type token can have the specific rules to check if token valid.
+   * See those classes for the validation logic specific to each type.
+   * @param token the token to validate
+   * @param context the context used for token validation
+   * @return future with token is validation was successful or failed future otherwise
+   */
+  public static Future<Token> validateTokenType(Token token, TokenValidationContext context) {
     return token.validateContext(context);
   }
```
steveellis, line 121: Not sure if this method adds any value. Why not just call `validateContext` directly?

SerhiiNosko: Because currently method validateContext is protected abstract, and I did not want to make it public. Also in case of applying this change need to do public of all 6 implementations

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
SerhiiNosko: Good catch, thank you.
I have deleted this method, now things have become simpler.

src/main/java/org/folio/auth/authtokenmodule/tokens/Token.java
steveellis: I see ok. You're not wanting to call `parse` a second time. That's fine. But how about just naming your new method `validate`. So `validateTokenType(Token token, TokenValidationContext context)` becomes `validate(Token token, TokenValidationContext context)`. Because I think here it's really just another signature of validation which can contain the token for convenience in case it has already been parsed..

SerhiiNosko: Agree, committed change

steveellis (approved):

PR #148, opened 2023-10-03
MODAT-151: Implement enhanced security mode

steveellis:
Please see the ticket: https://issues.folio.org/browse/MODAT-151

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,11 @@ mod-authtoken supports a number of command line options as system properties, se
 * `cache.permissions` - Boolean controlling the permissions cache. Defaults to `true`.
 * `allow.cross.tenant.requests` - Boolean to allow (in consortia setups) or deny cross tenant requests. Defaults to `false`.
 * `token.expiration.seconds` - Override defaults for token expiration in the form of `tenantId:<tenant id>,accessToken:<seconds>,refreshToken:<seconds>;accessToken:<seconds>,refreshToken:<seconds>`. To override defaults for a specific tenant provide a triplet. To override defaults provide a pair. Separate entries in the string with a `;` character. Neither tenant entries nor a default are required. If a default or a key is not provided, a default of 10 minutes is set by the module for the access token, and a default of one week is set by the module for the refresh token.
-
+* `enhnanced.security.tenants` - A comma separated list of tenant ids for which legacy tokens are not supported.
+*
 # Environment variables
-`TOKEN_EXPIRATION_SECONDS` - Identical to `token.expiration.seconds` as specified above. Provided as a convenience. System property takes precedence.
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/auth/authtokenmodule/tokens/legacy/EnhancedSecurityTenants.java
```diff
@@ -0,0 +1,35 @@
...
+    var prop = getTenantsFromEnvOrSystemProperty();
+    if (prop == null) {
+      return Collections.emptyList();
+    }
+    return Arrays.asList(prop.replace(" ", "").split(","));
+  }
+
+  private String getTenantsFromEnvOrSystemProperty() {
+    var prop = System.getProperty(ENHANCED_SECURITY_TENANTS);
+    if (prop != null) {
```
SerhiiNosko, line 30: better to use StringUtils.isNotBlank(prop) here, and the same in previous method, because someone can path empty string that is not valid case for us

steveellis: This is a good catch Serhii. I decided to check for the empty string and then throw an exception just in case people think that having the environment variable key is enough to enable enhanced security mode for all tenants. Although IDK maybe that should be a feature of this, to have a mechanism to enable it for all tenants...

SerhiiNosko (approved):

README.md
```diff
@@ -45,9 +45,11 @@ mod-authtoken supports a number of command line options as system properties, se
 * `cache.permissions` - Boolean controlling the permissions cache. Defaults to `true`.
 * `allow.cross.tenant.requests` - Boolean to allow (in consortia setups) or deny cross tenant requests. Defaults to `false`.
 * `token.expiration.seconds` - Override defaults for token expiration in the form of `tenantId:<tenant id>,accessToken:<seconds>,refreshToken:<seconds>;accessToken:<seconds>,refreshToken:<seconds>`. To override defaults for a specific tenant provide a triplet. To override defaults provide a pair. Separate entries in the string with a `;` character. Neither tenant entries nor a default are required. If a default or a key is not provided, a default of 10 minutes is set by the module for the access token, and a default of one week is set by the module for the refresh token.
-
+* `enhnanced.security.tenants` - A comma separated list of tenant ids for which legacy tokens are not supported.
```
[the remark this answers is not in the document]
steveellis: I do agree this is better. But what is the behavior if the property is null? All tenants are legacy tokens or none? None would be better IMO. Make legacy tokens in Poppy something that you have to explicitly enable. It's not that legacy tokens aren't supported (which is what the TC is worried about), just have to make a conscious decision to use them in operations.

PR #154, opened 2023-11-15
MODAT-160: When a token isn't present in a request, return 400 rather than 403

steveellis:
https://issues.folio.org/browse/MODAT-160

This doesn't change all uses of the `403` response code, only the one which is causing trouble for the front end developers. `403` is still returned when the user id in `X-Okapi` and the token don't match. I think it is fine to still have other cases that return `403`. The key here is that we no longer use it for missing tokens.


zburke:
This sounds good to me. 400 for "you did not present a token" makes sense to me since that's really an auth-n problem, whereas the 403 for "the token you presented was inadequate" is an auth-z problem, even though in both cases the result is the same: NO SOUP FOR YOU!

